Issue #81, opened 2021-12-20
wrong value of phase shows stack trace - should be handled gracefully


sonalgoyal:
try giving phase like lbel or something

navinrathore:
Reference for standard exit codes.
https://tldp.org/LDP/abs/html/exitcodes.html#EXITCODESREF

navinrathore:
Below is the Error message before code change
2021-12-22 15:02:23,176 [main] WARN  zingg.client.util.Email - Unable to send email Can't send command to SMTP host
 2021-12-22 15:02:23,176 [main] WARN  zingg.client.Client - Apologies for this message. Zingg has encountered an error. An error has occured while setting up the clientnull

PR #84, opened 2021-12-22
Graceful message and exit for wrong value of phase

navinrathore:
#81 wrong value of phase shows stack trace - should be handled gracefully

### DIFF
--- a/client/src/main/java/zingg/client/Client.java
+++ b/client/src/main/java/zingg/client/Client.java
@@ -8,6 +8,7 @@ import org.apache.spark.api.java.JavaSparkContext;
 
 import zingg.client.util.Email;
 import zingg.client.util.EmailBody;
+import zingg.client.util.Util;
 
 /**
  * This is the main point of interface with the Zingg matching product.
@@ -21,6 +22,7 @@ public class Client implements Serializable {
 	private ClientOptions options;
 
 	public static final Log LOG = LogFactory.getLog(Client.class);
+	public static final int EXIT_STATUS_ILLEGAL_CMD = 127;
 
 
 	/**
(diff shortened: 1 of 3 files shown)


sonalgoyal (commented):
Cant we just throw an exception and exit with the right message?

client/src/main/java/zingg/client/Client.java
```diff
@@ -21,6 +22,7 @@
 	private ClientOptions options;
 
 	public static final Log LOG = LogFactory.getLog(Client.class);
+	public static final int EXIT_STATUS_ILLEGAL_CMD = 127;
```
sonalgoyal, line 25: this is not needed

client/src/main/java/zingg/client/Client.java
```diff
@@ -131,6 +133,11 @@ public static void main(String... args) {
 				System.exit(0);
 			}
 			String phase = options.get(ClientOptions.PHASE).value.trim();
+			if (ZinggOptions.getByValue(phase) == null) {
```
sonalgoyal, line 136: throw exception with right message

client/src/test/java/zingg/client/TestClient.java
```diff
@@ -1,12 +1,66 @@
...
+
 	@Test
-	public void testInvalidArgs() {
-		
+	public void testInvalidPhaseArg() {
+		String s = null;
+		int exitStatusExpected = 127; // error code for an illegal command e.g. typo
+		try {
+			String zinggInvalidPhaseCmd = "scripts/zingg.sh --phase findTrainingDat --conf examples/febrl/config.json";
+			Process p = Runtime.getRuntime().exec(zinggInvalidPhaseCmd, null, new File(System.getenv("ZINGG_HOME")));
```
sonalgoyal, line 20: can we just call the client with the args instead of execing a new process here? and simply assert that an exception happens ?

sonalgoyal (changes requested):

client/src/main/java/zingg/client/Client.java
navinrathore: ok

client/src/main/java/zingg/client/Client.java
navinrathore: This is the message as part of thrown exception
 Apologies for this message. Zingg has encountered an error. 'findTrainingDat' is not a valid phase. Valid phases are: train|match|trainMatch|findTrainingData|label|link

client/src/test/java/zingg/client/TestClient.java
navinrathore: testcases have been updated to test the new functionality that is refactored into ZinggOptions.verifyPhase(phase).
ZinggOptions has been chosen for it is the central place to maintain zingg options or phases.

sonalgoyal (approved):

Issue #87, opened 2021-12-24
labeller should indicate how many pairs are matches and non matches


sonalgoyal:
right now we have no clue how many pairs we have marked and as what. A breakdown at every labelling question will help.

PR #94, opened 2021-12-28
Display how many pairs've been labelled matches and non matches #87

navinrathore:
During the Labeling phase, track the number of pairs that has been marked match and non match. Display same to the user on cli.


sonalgoyal (changes requested):
as discussed, cant say stats have to be there. stats have to be overall, not just existing round

navinrathore:
Consolidated stats are printed

sonalgoyal:
> as discussed, cant say stats have to be there. stats have to be overall, not just existing round

does this change have cant say stats as commented on previously?

sonalgoyal (changes requested):
i dont see cant say stats?

navinrathore:
Not Sure has been added

Issue #83, opened 2021-12-21
Log training data and test data locations


sonalgoyal:
It is difficult to understand from the logs where the training data is coming from - we see sometimes that it just prints no training data but we do not know where it is reading from. See #79

PR #97, opened 2021-12-29
logging location of the file being read #83

navinrathore:
The output location would look like : (2nd line)
2021-12-29 18:18:22,739 [main] WARN  zingg.util.PipeUtil - Reading input PARQUET
**2021-12-29 18:18:22,740 [main] WARN  zingg.util.PipeUtil - Reading location: models/1000/trainingData//unmarked/**

If entire pipe info is printed, for some scenario other information also get printed. So this option dropped
e.g.
Reading input Pipe [name=null, format=PARQUET, preprocessors=null, props={location=models/1000/trainingData//unmarked/}, schema=null]
Reading input Pipe [name=customers, format=SNOWFLAKE, preprocessors=null, props={sfUrl=tsa87485.snowflakecomputing.com, sfUser=sonalgoyal, sfPassword=ZZZZZ, sfDatabase=DEMO_DB, sfSchema=PUBLIC, sfWarehouse=COMPUTE_WH, dbtable=CUSTOMERS}, schema=null]

### DIFF
--- a/core/src/main/java/zingg/util/PipeUtil.java
+++ b/core/src/main/java/zingg/util/PipeUtil.java
@@ -69,6 +69,7 @@ public class PipeUtil {
 
 	private static Dataset<Row> read(DataFrameReader reader, Pipe p, boolean addSource) {
 		Dataset<Row> input = null;
+		LOG.warn("Reading " + p);
 		if (p.getProps().containsKey(FilePipe.LOCATION)) {
 			input = reader.load(p.get(FilePipe.LOCATION));
 		}


sonalgoyal:
We need the second option.

sonalgoyal (changes requested):
Need to print entire pipe. It will become an issue for others sensitive information like passwords etc, log an issue for that separately

navinrathore:
There are two place where the entire pipe information can be logged.
a) inside read(), same place as done in current commit. It is a new statement
b) update info in existing logging statement but a bit higher place in getReader() that is invariably executed before reading.
 PipeUtil.java
58  LOG.warn("Reading input " + p.getFormat().type());

Please suggest.

navinrathore:
> There are two place where the entire pipe information can be logged. a) inside read(), same place as done in current commit. It is a new statement b) update info in existing logging statement but a bit higher place in getReader() that is invariably executed before reading. PipeUtil.java 58 LOG.warn("Reading input " + p.getFormat().type());
> 
> Please suggest.

Option A has been chosen as advised.

sonalgoyal (changes requested):
seems there is some confusion,. We need to print the entire pipe not just location.

sonalgoyal:
> Need to print entire pipe. It will become an issue for others sensitive information like passwords etc, log an issue for that separately

this with the right read method

sonalgoyal (changes requested):
https://github.com/zinggAI/zingg/pull/97#issuecomment-1002658799

navinrathore:
Logging of pipe information is done in all cases

Issue #58, opened 2021-12-01
Ability to update labels once marked 


sonalgoyal:
Right now there is no way to update labels if a wrong one has been specified. See https://zinggai.slack.com/archives/C02JNH144TB/p1638361934009800

sonalgoyal:
We need to spec this out.

sonalgoyal:
What if we had an html containing all the pairs and their labels? The label phase could generate this with the round number and pair information everytime. 

we could then define a new phase - updateLabel  which takes in the id/ids of the pairs we want to update

sonalgoyal:
Let us make a folder html under model/modelId and have same marked/unmaked. Write html version of the dataframe to this folder after findtrainingData and label phases.

sonalgoyal:
This html framework will eventually be used for dq, stats, model documentation etc so we have to write it in a generic way

sonalgoyal:
After discussion with @navinrathore :
- Separate phase generateTrainingDataDocs which goes over marked and unmarked records and builds out the html. The html contains pagination and search. And separate sections for records - marked (match/non match/cant say) and unmarked 
- Focus on readability of pairs in all cases 
- Write under model/modelId/docs/trainingData.html with subfiles.

sonalgoyal:
Have a phase updateLabel which first invokes generateTrainingDataDocs and asks the user to tell which record id to update in a loop. It then updates the labels and generates the docs again

sonalgoyal:
> Have a phase updateLabel which first invokes generateTrainingDataDocs and asks the user to tell which record id to update in a loop. It then updates the labels and generates the docs again

@navinrathore can you please work on this?

sonalgoyal:
Have a phase updateLabel and asks the user to tell which record id to update in a loop. It then updates the labels and generates the docs again

PR #102, opened 2021-12-31
new phase updateLabel implementation #58

navinrathore:
Implementation of LabelUpdater extended from labeller.
Output/Prints are similar in both the phases
Marked Records are replaced

Local Branch has been fetched from 0.3.1

### DIFF
--- a/core/src/main/java/zingg/LabelUpdater.java
+++ b/core/src/main/java/zingg/LabelUpdater.java
@@ -1,0 +1,129 @@
+package zingg;
+
+import java.util.List;
+import java.util.Scanner;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.spark.sql.Column;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SaveMode;
+
+import zingg.client.ZinggClientException;
+import zingg.client.ZinggOptions;
+import zingg.client.pipe.Pipe;
+import zingg.client.util.ColName;
+import zingg.client.util.ColValues;
+import zingg.util.DSUtil;
+import zingg.util.PipeUtil;
+
+public class LabelUpdater extends Labeller {
+	protected static String name = "zingg.LabelUpdater";
+	public static final Log LOG = LogFactory.getLog(LabelUpdater.class);
+
+	public LabelUpdater() {
+		setZinggOptions(ZinggOptions.UPDATE_LABEL);
+	}
+
+	public void execute() throws ZinggClientException {
+		try {
+			LOG.info("Reading inputs for updateLabelling phase ...");
+			Dataset<Row> markedRecords = PipeUtil.read(spark, false, false, PipeUtil.getTrainingDataMarkedPipe(args));
+			processRecordsCli(markedRecords);
+			LOG.info("Finished updataLabelling phase");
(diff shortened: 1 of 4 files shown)


core/src/main/java/zingg/LabelUpdater.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class LabelUpdater extends Labeller {
+	protected static String name = "zingg.LabelUpdater";
+	public static final Log LOG = LogFactory.getLog(LabelUpdater.class);
+
+	public LabelUpdater() {
+		setZinggOptions(ZinggOptions.UPDATE_LABEL);
+	}
+
+	public void execute() throws ZinggClientException {
```
sonalgoyal, line 29: rename getUnmarkedRecords in labeller to getRecordsToProcess, call that in execute. In this class, override that method so that you do not have to override execute here

core/src/main/java/zingg/LabelUpdater.java
```diff
@@ -0,0 +1,129 @@
...
+			if (LOG.isDebugEnabled()) {
+				e.printStackTrace();
+			}
+			LOG.warn("An error has occured while Updating Label. " + e.getMessage());
+			throw new ZinggClientException(e.getMessage());
+		}
+		return;
+	}
+
+	private void updateLabellerStat(int selectedOption, int existingType) {
```
sonalgoyal, line 106: have one method updateLabellerstat(selectedOption, existingType, increment) in base class and call that with increment +1 there or -1 here

core/src/main/java/zingg/LabelUpdater.java
```diff
@@ -0,0 +1,129 @@
...
+		else if (existingType == ColValues.MATCH_TYPE_NOT_A_MATCH) {
+			--negativePairsCount;
+		}
+		else if (existingType == ColValues.MATCH_TYPE_NOT_SURE) {
+			--notSurePairsCount;
+		}
+		updateLabellerStat(selectedOption);
+	}
+
+	void writeLabelledOutput(Dataset<Row> records, SaveMode mode) {
```
sonalgoyal, line 120: define method getPipeForRecordsTowrite in base class. override it here with SaveMode. call it in writeLAbelledOutput in base class.

sonalgoyal (changes requested):
Comments on overriding classes and methods

core/src/main/java/zingg/LabelUpdater.java
```diff
@@ -0,0 +1,129 @@
...
+	public void processRecordsCli(Dataset<Row> lines) throws ZinggClientException {
+		LOG.info("Processing Records for CLI updateLabelling");
+		getMarkedRecordsStat(lines);
+		printMarkedRecordsStat();
+		if (lines == null || lines.count() == 0) {
+			LOG.info("There is no marked record for updating. Please run findTrainingData/label jobs to generate training data.");
+			return;
+		}
+
+		List<Column> displayCols = DSUtil.getFieldDefColumns(lines, args, false);
```
sonalgoyal, line 50: displaycols is redundant, we use df.show so i doubt if it being used. please check

sonalgoyal: also, overall is there a way to abstract the logic - reading and quitting at 9, displaying some stuff, accepting user input, doing some action(update/mark) in the base class and only override the relevant methods in the update labeller class?

Issue #108, opened 2022-01-04
When linking 3+ datasets, all records that match against the master source should have same z_cluster


danieldynasty:
All records that match against the master source should have same z_cluster. The output is currently in pairs.

sonalgoyal:
Thanks for reporting this, it makes sense to align the output with same z_cluster from the master source. Will fix!

navinrathore:
For Linking phase only, cluster id could be comprised of z_id. In this case, z_id of source[0]. Thereafter, all the related records will automatically have unique cluster_id.

Else, set of records with same z_id shall be given some other unique identifier.

PR #109, opened 2022-01-05
Giving same cluster id to all records linked from multiple sources #108

navinrathore:
addUniqueCol() is also referred to TrainingDataFinder. 
dropDuplicate() could be expansive, but perhaps necessary. Applied on data from master source only.

Note: addUniqueCol() is also duplicated in FileUtil.java, nowhere used as of now.


sonalgoyal (changes requested):
Instead of changing trainingdatafinder and addUnqieCol, can we invoke addUniqueCol with id_col and change alignLinked method to make z_cluster same as z_id ?

navinrathore:
Yes. That's look good. Made the changes.

Issue #95, opened 2021-12-28
Support for float type


doku:
Support for float or double type for matching. Columns with decimal numbers can't be used for matching right now. It would be useful if these numbers can be used to match within a few +- percentage.

sonalgoyal:
@doku are you able to share some data which we can use for testing this?

mliscombe:
Relevant data here:
https://github.com/anhaidgroup/deepmatcher/blob/master/Datasets.md
http://pages.cs.wisc.edu/~anhai/data1/deepmatcher_data/Structured.zip

sonalgoyal:
Thanks 🙏🏽 @mliscombe

navinrathore:
With the above data, we have confirmed that support of float/double and integer data is already present in the Zingg.
However, we have found a bug that would restrict working end-to-end with such data.
We'll fix the same under this issue.

navinrathore:
The "z_source" column should be added at the end of the record for consistency. The alignDupes() function requires desired change.
The alignLinked() may also require similar change.
The other place of same interest is getFieldDefColumns(). All functions are from DSUtil.java.

PR #119, opened 2022-01-08
z_source column is added at the end of dataset in alignDupes 

navinrathore:
It is fine to have z_source in the beginning in alignLinked() for better user experience. It's an end output file.
This file is mandatory for the newly generated models to test.
Reference GH-95


sonalgoyal (changes requested):
alignLinked to change too to make it consistent with the match output. Or we can have all the z columns in the beginning. that may be a bigger change for now. so lets change alignlinked and we can revisit later

Issue #82, opened 2021-12-20
Blocking tree can not be saved in cloud environment


sonalgoyal:
the blocking tree gets written using local file system apis which do not work in cloud env like databricks. If we do trainMatch in one go, the file probably goes into the local machine form where it will get picked up. Otherwise it will not work if we do train and match separately.

sonalgoyal:
can you please check if there is a way to write the serialized blockingtree to parquet using spark.write? We may(?) need to coalesce to 1 to make sure only one file is written. Then we can read like standard spark files without worrying about the location being local. @navinrathore

navinrathore:
[From the typical datasources used in Spark](https://spark.apache.org/docs/latest/sql-data-sources-parquet.html), The solution mentioned https://stackoverflow.com/questions/35200988/writing-custom-java-objects-to-parquet looks promising. but it will need additional dependency jar (org.apache.parquet  (parquet-avro.jar))For this or other format, we need schema of the object. It can get the same from class object of (Tree<Canopy>). But I am getting runtime issue to get the same, thoughPlease take a glance on the approach, we can discuss it then. or let me know if you have any suggestion
Stack OverflowStack Overflow
Writing custom java objects to Parquet
I have some custom java objects (which internally are composed of other custom objects). I wish to write these to HDFS in parquet format. Even after a lot of searching, most suggestions seem to be
2:48
Seems the getScehme() api here does not work for generic type. Our type is Tree<Canopy>
2:50
Can we have  this schema defined separately?

navinrathore:
>>>Can’t we serialise the bytes?
>>>Serealise object to bytearray and write that
>>>Then read and reconstruct

We can only read binary_file. reading not supported

sonalgoyal:
I don’t have a ready answer here but I think one of the following approaches should work.
A. Write object as bytes
B. Write as json
C. Build spark dataset with canopy class and read and write through that

in all cases, we should invoke the spark apis for reading and writing. So essentially serialise as byte/json/custom obj and build a dataset and use spark.read or spark.write

navinrathore:
Option A (object in bytes) has worked. PR #120

PR #120, opened 2022-01-08
Blocking tree are saved in parquet file #82

navinrathore:
"zingg.block" is a directory comprising parquet files. It was a file.
BlockingtreePipe, by default, has Overwrite mode. It should be OK in read().

### DIFF
--- a/client/src/main/java/zingg/client/util/Util.java
+++ b/client/src/main/java/zingg/client/util/Util.java
@@ -1,5 +1,7 @@
 package zingg.client.util;
 
+import java.io.BufferedInputStream;
+import java.io.BufferedOutputStream;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
@@ -473,7 +475,8 @@ public class Util implements Serializable {
 	
 	public static byte[] convertObjectIntoByteArray(Object obj) throws IOException {
 		ByteArrayOutputStream bos = new ByteArrayOutputStream();
-		ObjectOutputStream oos = new ObjectOutputStream(bos);
+		BufferedOutputStream bufferedOS = new BufferedOutputStream(bos);
+		ObjectOutputStream oos = new ObjectOutputStream(bufferedOS);
 		oos.writeObject(obj);
 		oos.flush();
 		return bos.toByteArray();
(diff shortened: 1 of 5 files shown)


client/src/main/java/zingg/client/util/Util.java
```diff
@@ -469,7 +471,18 @@ public static String getCurrentLabelPath(String path) throws IOException {
 		return dupesActual;
 	}
 	
-	
-
```
sonalgoyal, line 473: should there be some kind of buffering? bufferedoutputstream etc?

core/src/main/java/zingg/util/BlockingTreeUtil.java
```diff
@@ -69,6 +76,21 @@
...
 	
-	
-    
+	public static void writeBlockingTree(SparkSession spark, JavaSparkContext ctx, Tree<Canopy> blockingTree, Arguments args) throws Exception {
+		byte[] byteArray  = Util.convertObjectIntoByteArray(blockingTree);
+		StructType schema = DataTypes.createStructType(new StructField[] { DataTypes.createStructField("BlockingTree", DataTypes.BinaryType, false) });
+		List<Object> objList = new ArrayList<>();
+		objList.add(byteArray);
+		JavaRDD<Row> rowRDD = ctx.parallelize(objList).map((Object row) -> RowFactory.create(row));
+		Dataset<Row> df = spark.sqlContext().createDataFrame(rowRDD, schema).toDF();
```
sonalgoyal, line 85: shouldnt we coalesce the df before writing?

sonalgoyal (changes requested):
please take a look at the comments

client/src/main/java/zingg/client/util/Util.java
navinrathore: There should not be any buffering, I feel. Buffering happens mainly for relatively expensive operations like disk I/O, network activity. Also, particular operation should be quite often.
Your opinion please?

core/src/main/java/zingg/util/BlockingTreeUtil.java
navinrathore: Added.

client/src/main/java/zingg/client/util/Util.java
sonalgoyal: buffering is pretty standard, dont see how it harms. please add.

sonalgoyal (changes requested):
buffering

client/src/main/java/zingg/client/util/Util.java
navinrathore: Yes. Added. Hope it is the correct way and does the expected.

PR #121, opened 2022-01-08
new febrl models

navinrathore:
The febrl models have been updated after
a) changes in config files - areacode, 'exact' for street no, areacode 
b) z_source movement at the end of dataset  (#95)
c) Blocking tree read/write through spark (#82)


examples/febrl/configWithTrainingSamples.json
```diff
@@ -120,7 +128,7 @@
 		}],
 		"labelDataSampleSize" : 0.5,
 		"numPartitions":4,
-		"modelId": 100,
-		"zinggDir": "models"
-
+		"modelId": 500,
+		"zinggDir": "models",
```
sonalgoyal, line 132: our model ids should be 100 series. also add a modelId.txt file under examples with id and description of models.

sonalgoyal (changes requested):

Issue #114, opened 2022-01-07
Labeller - if it is not known that it is a match or not, similarity score should not be printed


sonalgoyal:
Current Message:
Zingg predicts the above records ARE NOT KNOWN IF MATCH with a similarity score of 0.00

New Message(only if not known)
Zingg does not any prediction for the above pairs as Zingg is still collecting training data to build the preliminary models.

navinrathore:
Please elaborate bit more on the scenario.
working with febrl data, I haven't found above case. all are either MATCH or NON MATCH (zingg prediction)

sonalgoyal:
The first time we run findTrainingData and label, them labelled prints this

navinrathore:
if prediciton is IS_NOT_KNOWN_PREDICTION, the new message should be printed.

sonalgoyal:
yes, thats probably one way of looking at it.

PR #125, opened 2022-01-10
Labeller - if it is not known that it is a match or not, similarity score should not be printed #114

navinrathore:
Improved message if prediction is not known.
It also has collectMetrics CLI option changes.
Unimportant messages have been removed for collectMetrics case.


navinrathore:
#114

Issue #124, opened 2022-01-10
trainingSamples need a fix


sonalgoyal:
we should either update the training samples with a z_source column or add that if not found, else findTrainingData/label breaks with pre existing training data

sonalgoyal:
lets figure out the best way to use unionByName if that works @navinrathore

sonalgoyal:
2022-01-12 18:40:51,773 [main] WARN  zingg.util.PipeUtil - Reading Pipe [name=trainingPos, format=CSV, preprocessors=null, props={location=examples/febrl/training.csv, delimiter=,, header=false, badRecordsPath=/tmp/bad}, schema=StructType(StructField(z_cluster,StringType,false), StructField(z_ismatch,IntegerType,true), StructField(fname,StringType,true), StructField(lname,StringType,true), StructField(stNo,StringType,true), StructField(add1,StringType,true), StructField(add2,StringType,true), StructField(city,StringType,true), StructField(areacode,StringType,true), StructField(state,StringType,true), StructField(dob,StringType,true), StructField(ssn,StringType,true))]
 2022-01-12 18:40:52,027 [main] WARN  zingg.util.DSUtil - Read all training samples
 2022-01-12 18:41:02,222 [main] WARN  zingg.client.util.Email - Unable to send email null
 2022-01-12 18:41:02,223 [main] WARN  zingg.client.Client - Apologies for this message. Zingg has encountered an error. org.apache.spark.sql.Dataset.unionByName(Lorg/apache/spark/sql/Dataset;Z)Lorg/apache/spark/sql/Dataset;
 
 
 steps 
/scripts/zingg.sh --phase findTrainingData --conf examples/febrl/configWithTrainingSamples.json
 ./scripts/zingg.sh --phase label --conf examples/febrl/configWithTrainingSamples.json
 ./scripts/zingg.sh --phase findTrainingData --conf examples/febrl/configWithTrainingSamples.json

sonalgoyal:
this was a spark home error at my end.

delgermurun:
Nice.

@sonalgoyal when will it be released along with a new Docker image?

PR #127, opened 2022-01-12
unionByName() with allowMissingColumns=true for training data

navinrathore:
added z_source, removed id from training.csv #124
z_source has been named as "training"
All the missing fields from either DS take null value
Screenshot from 2022-01-12 06-24-58

sample records after union
Screenshot from 2022-01-12 07-06-29


navinrathore:
z_source is not included in training samples.

Issue #110, opened 2022-01-05
More example models


sonalgoyal:
Relevant data here:
https://github.com/anhaidgroup/deepmatcher/blob/master/Datasets.md
http://pages.cs.wisc.edu/~anhai/data1/deepmatcher_data/Structured.zip

_Originally posted by @mliscombe in https://github.com/zinggAI/zingg/issues/95#issuecomment-1005369119_

navinrathore:
To work with next 2 datasets namely
1. BeerAdvo-RateBeer
2. iTunes-Amazon1

PR #144, opened 2022-02-15
New Datasets iTunes-amazon, beerAdvo-rateBeer and respective models

navinrathore:
$ sign from Price and % sign from ABV have been removed. (README.txt files contains details)
Both fields are of  type DoubleType
All other fields are of type StringType and  "fuzzy" matchType is used


navinrathore:
Tracking issue #110

Issue #134, opened 2022-01-18
Interactive Labeller - Does not match at probability 0.5


sonalgoyal:
During model building of 5M NC dataset, saw some messages on the interactive learner - what does do not match with probability 0.5 mean? 
![Screenshot from 2022-01-14 20-55-06](https://user-images.githubusercontent.com/309387/149877111-238ab985-254d-4ad0-9075-fed049dd3bd8.png)

navinrathore:
The typical examples of the concerned statement are (with complete score value)
================================================================
Zingg predicts the above records DO NOT MATCH with a similarity score of 0.000000
Zingg predicts the above records DO NOT MATCH with a similarity score of 0.361295
Zingg predicts the above records DO NOT MATCH with a similarity score of 0.486515
Zingg predicts the above records DO NOT MATCH with a similarity score of 0.499993
Zingg predicts the above records MATCH with a similarity score of 1.000000
Zingg predicts the above records MATCH with a similarity score of 0.518476
Zingg predicts the above records MATCH with a similarity score of 0.999999
================================================================
As, currently in code, the precision for the "similarity score" is set to two decimal points. the value is rounded. e.g, 0.499993 gets converted to 0.50. Hence, on console, the message says "... above records DO NOT MATCH with a similarity score of 0.50".

sonalgoyal:
please make it a % score to 2 dec places  then

sonalgoyal:
or rather do this https://stackoverflow.com/questions/1305827/prevent-round-off-in-string-format-2f-doublevalue-in-java?

PR #145, opened 2022-02-15
prediction score: set precision to 2 with ~ floor mode

navinrathore:
It works like below.
e.g.
MATCH with a similarity score of {score:%f} {score:%.2f}  {newScore:%.2f}
MATCH with a similarity score of 0.965285 0.97 0.96
MATCH with a similarity score of 0.999966 1.00 0.99


navinrathore:
issue #134

PR #149, opened 2022-02-25
In TrainMatcher class, ZinggOptions set to TRAIN_MATCH

navinrathore:
ZinggOptions set to correct value i.e. "TRAIN_MATCH" for the class
This change gives correct google analytics data.


navinrathore:
#149

PR #152, opened 2022-02-26
Initial changes for Spark to Snowpark migration

navinrathore:
Few Points:
Compiler version - java 11 (Fixed)
Scala version 2.12.11 and above

* simple df.col(x) is not present. Also no Columns().... in snowspark, it has to get something like df.schema().names() ...  need to work further on this.
* collectAsList() is not present. rather, need to use collect(): Seq and get the needful done.
* read and write are simpler. Perhaps, most of the time have to be writeToTable() for core data.

### DIFF
--- a/client/src/main/java/zingg/client/pipe/Pipe.java
+++ b/client/src/main/java/zingg/client/pipe/Pipe.java
@@ -5,11 +5,15 @@ import java.io.Serializable;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


sonalgoyal:
can you please delete this pr and send me one pr which has pom.xml, trainingdatafinder and pom.xml ? The zinggBase.java and PipeUtil - i may have to work on them. In the mean time, can you please add another option --snowflake as above?

PR #157, opened 2022-03-01
snowpark migration: basic APIs replacement

navinrathore:
* drop() and sort() do not work with vararg type of inputs. However, null gets removed as drop(Column[]), sort(Column[]) versions used
* drop(), sort(), asc(), desc(), schema(), sample(), rename() tested in a sample application
* compilation errors due to signature mismatch and dependency on other classes  
    - Block
    - BlockingTreeUtil
    - RowEncoder
    - StorageLevel
   -  read/write
   -  DSUTil couple of functions signature mismatch (Dataset to DataFrame)
* func(Columns[]) version prefereed over func(String[]) version e.g. sort(), drop()
* $eq$eq$eq   ===,  <=> are also the same

### DIFF
--- a/client/src/main/java/zingg/client/util/Util.java
+++ b/client/src/main/java/zingg/client/util/Util.java
@@ -22,24 +22,18 @@ import java.util.Scanner;
 import java.util.Set;
 import java.util.regex.Pattern;
 
+import com.snowflake.snowpark.Column;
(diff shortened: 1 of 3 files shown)


navinrathore:
This PR is through another branch. It takes care of all the comments mentioned in that PR.

Issue #159, opened 2022-03-03
Add user to docker command


jdicello:
**Describe the bug**
The current docker command does not specify a user.  When the container is run w/o a user, you are unable to modify the file system. The instructions ask you to rename a directory, but that command will fail.  Additionally, VS Code requires the ability to create directories to connect to a running container.

**To Reproduce**
1. Run the commands:
docker pull zingg/zingg:0.3.2
docker run -it zingg/zingg:0.3.2 bash
3. When the session has started, type the command "mkdir test_dir"
![image](https://user-images.githubusercontent.com/2998073/156631610-e46005a4-93a7-40c8-affb-c2a8f8a41643.png)

**Expected behavior**
The directory "test_dir" should be created.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] WSL2 - Ubuntu

**Smartphone (please complete the following information):**

**Additional context**
I was able to fix the problem by adding "-u root" to the docker command.  I am not sure if this is the ideal way to give this permission.

sonalgoyal:
Thanks for reporting this @jdicello. we will try to fix this as soon as possible. @navinrathore can you please take a look at this urgently? Thanks.

sonalgoyal:
Added docker instructions https://docs.zingg.ai/zingg/stepbystep/installation/workingwithdocker. Hope this helps @jdicello

sonalgoyal:
Confirmed with @jdicello and closing

PR #160, opened 2022-03-04
Documentaion for Working with Docker Image

navinrathore:
Changed permission to rw the directories


navinrathore:
#159

navinrathore:
User can create directories in any of the sub directories of /zingg-0.3.2-SNAPSHO.
The cp, mv and other basic commands were already working.

sonalgoyal (changes requested):
can we have a new folder /zinggData and /zinggModels and make them writable? rw on /zingg can be susceptible to installation going wrong

Also, we will need a new release to push this. Can you make appropriate documentation, pom Client.java and other changes?

PR #164, opened 2022-03-07
 snowflake_java based apis

navinrathore:
Replaced with snowflake_java api. Cleaner code.  Now have vararg (Column ..) based functions

However, below are the few things which are need to be resolveed

a) filter() -> as arg two columns are compared by equalTo(Column) ..
    earlier used $eq$eq$eq,  but in java api no equal_to(Object) function
    .in(Object) is the nearest one but there is no not_in(Object) function
b) Row().getAs("ColumnName") fn is not there rather Row().get(Index) is present....Need to get convert from column index
c) DF.Util()    GraphFrame is spark based
d) persist(), rapartition(0 etc. are WIP for snowpark

### DIFF
--- a/core/src/main/java/zingg/Labeller.java
+++ b/core/src/main/java/zingg/Labeller.java
@@ -7,6 +7,7 @@ import java.util.Scanner;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+
 import com.snowflake.snowpark_java.Column;
 import com.snowflake.snowpark_java.DataFrame;
 import com.snowflake.snowpark_java.Row;
@@ -57,7 +58,7 @@ public class Labeller extends ZinggBase {
 			}
 			if (markedRecords != null ) {
 				unmarkedRecords = unmarkedRecords.join(markedRecords,
-						unmarkedRecords.col(ColName.CLUSTER_COLUMN).in(markedRecords.col(ColName.CLUSTER_COLUMN)),
+						unmarkedRecords.col(ColName.CLUSTER_COLUMN).equal_to(markedRecords.col(ColName.CLUSTER_COLUMN)),
 						"left_anti");
 						getMarkedRecordsStat(markedRecords);
 			} 
@@ -68,9 +69,9 @@ public class Labeller extends ZinggBase {
 	}
 
 	protected void getMarkedRecordsStat(DataFrame markedRecords) {
-		positivePairsCount = markedRecords.filter(markedRecords.col(ColName.MATCH_FLAG_COL).in(ColValues.MATCH_TYPE_MATCH)).count() / 2;
-		negativePairsCount = markedRecords.filter(markedRecords.col(ColName.MATCH_FLAG_COL).in(ColValues.MATCH_TYPE_NOT_A_MATCH)).count() / 2;
-		notSurePairsCount = markedRecords.filter(markedRecords.col(ColName.MATCH_FLAG_COL).in(ColValues.MATCH_TYPE_NOT_SURE)).count() / 2;
+		positivePairsCount = markedRecords.filter(markedRecords.col(ColName.MATCH_FLAG_COL).equal_to(Functions.lit(ColValues.MATCH_TYPE_MATCH))).count() / 2;
+		negativePairsCount = markedRecords.filter(markedRecords.col(ColName.MATCH_FLAG_COL).equal_to(Functions.lit(ColValues.MATCH_TYPE_NOT_A_MATCH))).count() / 2;
+		notSurePairsCount = markedRecords.filter(markedRecords.col(ColName.MATCH_FLAG_COL).equal_to(Functions.lit(ColValues.MATCH_TYPE_NOT_SURE))).count() / 2;
 		totalCount = markedRecords.count() / 2;
 	}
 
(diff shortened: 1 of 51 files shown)


core/src/main/java/zingg/Labeller.java
```diff
@@ -95,11 +96,11 @@ public void processRecordsCli(DataFrame lines) throws ZinggClientException {
 			int totalPairs = clusterIDs.size();
 			
 			for (int index = 0; index < totalPairs; index++){	
-				DataFrame currentPair = lines.filter(lines.col(ColName.CLUSTER_COLUMN).in(
-						clusterIDs.get(index).getAs(ColName.CLUSTER_COLUMN))).cacheResult();
+				DataFrame currentPair = lines.filter(lines.col(ColName.CLUSTER_COLUMN).equal_to(
+					Functions.lit(clusterIDs.get(index).get(0)))).cacheResult();
```
sonalgoyal, line 100: We should refer to column names only, let us not use index for retrieval as it is very brittle

core/src/main/java/zingg/Trainer.java
```diff
@@ -35,26 +35,26 @@ public void execute() throws ZinggClientException {
...
+			DataFrame positives = null;
+			DataFrame negatives = null;
+			DataFrame tra = DSUtil.getTraining(snow, args);
 			tra = DSUtil.joinWithItself(tra, ColName.CLUSTER_COLUMN, true);
-			tra = tra.cache();
-			positives = tra.filter(tra.col(ColName.MATCH_FLAG_COL).equalTo(ColValues.MATCH_TYPE_MATCH));
-			negatives = tra.filter(tra.col(ColName.MATCH_FLAG_COL).equalTo(ColValues.MATCH_TYPE_NOT_A_MATCH));
+			tra = tra.cacheResult();
+			positives = tra.filter(tra.col(ColName.MATCH_FLAG_COL).in(ColValues.MATCH_TYPE_MATCH));
+			negatives = tra.filter(tra.col(ColName.MATCH_FLAG_COL).in(ColValues.MATCH_TYPE_NOT_A_MATCH));
```
sonalgoyal, line 44: why not functions.lit ?

core/src/main/java/zingg/hash/First3CharsBox.java
```diff
@@ -37,8 +37,8 @@ public Integer call(String field) {
 	 }
 	 
 	 @Override
-	 public Object apply(Row ds, String column) {
-		 return call((String) ds.getAs(column));
+	 public Object apply(Row ds, int column) {
+		 return call((String) ds.get(column));
```
sonalgoyal, line 41: let us discuss this

core/src/main/java/zingg/hash/IsNullOrEmpty.java
```diff
@@ -15,7 +15,7 @@ public Boolean call(String field) {
 		 return (field == null || ((String ) field).trim().length() == 0);
 	 }
 
-	public Object apply(Row ds, String column) {
-		 return call((String) ds.getAs(column));
+	public Object apply(Row ds, int column) {
+		 return call((String) ds.get(column));
```
sonalgoyal, line 19: to discuss

core/src/main/java/zingg/util/HashUtil.java
```diff
@@ -37,7 +38,7 @@
 				});
 		for (HashFnFromConf scriptArg : scriptArgs) {
 			HashFunction fn = HashFunctionRegistry.getFunction(scriptArg.getName());
-			spark.udf().register(fn.getName(), (UDF1) fn, fn.getReturnType());
+			snow.udf().registerPermanent(fn.getName(), (JavaUDF1<Vector, Double>) fn, fn.getDataType(), fn.getReturnType(), "stageLocation");
```
sonalgoyal, line 41: parameter type of JavaUDF - why is it a vector?

core/src/main/java/zingg/util/Metric.java
```diff
@@ -16,7 +16,7 @@
     public static final long timeout = 1200L;
     public static final double confidence = 0.95; // default value
 
-    public static double approxCount(Dataset<Row> data) {
-        return data.rdd().countApprox(timeout, confidence).initialValue().mean();
+    public static double approxCount(DataFrame data) {
+        return data.async().count().getResult((int)timeout/1000);
```
sonalgoyal, line 20: please comment this out

sonalgoyal (changes requested):

core/src/main/java/zingg/util/Metric.java
navinrathore: Excluded through pom.xml
Comment this function as well

core/src/main/java/zingg/util/HashUtil.java
navinrathore: remove parameters

core/src/main/java/zingg/Trainer.java
navinrathore: done

core/src/main/java/zingg/Labeller.java
navinrathore: changed in the best manner. Will revisit if we have general solution of Row elements extraction.

PR #178, opened 2022-03-20
Correctly handling Pipe initialization

edmondo1984:
When looking at the current code, there are some patterns which should be avoided:
- Mutable classes leads to objects in an undesirable state. See Joshua Bloch, Effective Java book, "Prefer Immutability"
- When an object construction phase is multi-step, the logic should be encapsulated in a builder (see https://blogs.oracle.com/javamagazine/post/exploring-joshua-blochs-builder-design-pattern-in-java)
- Throw exceptions that are relevant to the level of abstractions you are dealing with and let the caller handle that (see my change in PipeFactory)

This is still a DRAFT, but I wanted to share the approach before moving forward

### DIFF
--- a/client/src/main/java/zingg/client/pipe/FilePipe.java
+++ b/client/src/main/java/zingg/client/pipe/FilePipe.java
@@ -1,13 +1,41 @@
 package zingg.client.pipe;
 
+
+import org.apache.spark.sql.SaveMode;
+import org.apache.spark.sql.types.StructType;
+
+import java.util.Map;
+
 public class FilePipe extends Pipe{
 	
 	public static final String LOCATION = "location";
-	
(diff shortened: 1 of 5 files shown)


sonalgoyal (commented):
Thanks @edmondo1984 - looks super cool.  I would suggest running the code on a Spark cluster as we had to put the clone method etc to overcome some issues we saw at that time.

edmondo1984:
@sonalgoyal Sure. Do you see blocks when trying to execute tests automatically at every commit? This should allow to test the code using a spark single cluster

sonalgoyal:
No blockers as such @edmondo1984 - just that our unit testing is really something we need to focus a lot on but havent been able to. :-(

Issue #78, opened 2021-12-15
Automatically suggest stop words


sonalgoyal:
we should build pre precessing routines which automatically find a list of stop words per column which the user can edit

sonalgoyal:
--phase suggestStopWords --conf <json> --field <name of field from conf> --num <count> --output </tmp..>

this will do word count of words in the field and write out the top <count> words as output as a csv with one word per line.

PR #186, opened 2022-03-23
Extract Stop words

navinrathore:
* Extracting stop words for each DONOT_USE field.
* stopWords are saved into csv files and html docs. The html docs have been linked in base documentation model.html
* Added new config param - **stopWordsCutoff** (default value = 0.1)


navinrathore:
#78

client/src/main/java/zingg/client/Arguments.java
```diff
@@ -507,7 +508,7 @@ public String getZinggBaseModelDir() {
 
 	@JsonIgnore
 	public String getZinggDocDir() {
-		return getZinggBaseModelDir();
+		return zinggDir + "/" + modelId;
```
sonalgoyal, line 511: why do we need this change?

sonalgoyal (changes requested):

client/src/main/java/zingg/client/Arguments.java
navinrathore: getZinggDocDir() was unused.
getZinggBaseModelDir() is models/<id>/model directory. it contains model detail. 
The getZinggDocDir() seems corresponding to doc/documentation, therefore used after modification.

sonalgoyal: thanks. so cant we then keep returning the getzinggBaseModelDir which is anyways what you are returning?

PR #191, opened 2022-03-30
Data processing for Stop Words removal

navinrathore:


core/src/main/java/zingg/util/DSUtil.java
```diff
@@ -253,4 +262,67 @@
...
 	}
+
+	public static Dataset<Row> preprocessForStopWords(SparkSession spark, Arguments args, Dataset<Row> ds) {
+
+		List<String> wordList = new ArrayList<String>();
+		Dataset<Row> modDS = ds;
+		for (FieldDefinition def : args.getFieldDefinition()) {
+			if (!(def.getStopWords() == null || def.getStopWords() == "")) {
+				Dataset<Row> stopWords = PipeUtil.read(spark, false, false, PipeUtil.getStopWordsPipe(args, def.getStopWords()));
+				wordList = stopWords.toJavaRDD().map(new Function<Row, String>() {
```
sonalgoyal, line 273: do you need to convert to javardd here, cant we select the column _c0 and collect?

core/src/main/java/zingg/util/DSUtil.java
```diff
@@ -253,4 +262,67 @@
...
+				}).collect();
+
+				modDS = modDS.withColumn(def.getFieldName(), removeStopWords(wordList).apply(modDS.col(def.getFieldName())));
+			}
+		}
+
+		return modDS;
+	}
+
+	public static UserDefinedFunction removeStopWords(List<String> stopWords) {
```
sonalgoyal, line 286: move to new package. preprocess under core and have top level class

sonalgoyal: please also write junit for this

core/src/main/java/zingg/TrainingDataFinder.java
```diff
@@ -131,13 +141,16 @@ public void execute() throws ZinggClientException {
 			}	
     }
 
-	public void writeUncertain(Dataset<Row> dupesActual) {
+	public void writeUncertain(Dataset<Row> dupesActual, Dataset<Row> sampleOrginal) {
+		//dupesActual.show(4);
 		//input dupes are pairs
 		dupesActual = DFUtil.addClusterRowNumber(dupesActual, spark);
```
sonalgoyal, line 147: refer what we discussed today. alignDupes should have the actual values - please use join. select only relevant z_columns from dupesActual, join with sampleOriginal and pass to alignDupes.

sonalgoyal (changes requested):

core/src/main/java/zingg/util/DSUtil.java
navinrathore: List<string> is needed. select() and then collect() will give List<Row> or Array[ROW].
Still could it be simplified?

core/src/main/java/zingg/TrainingDataFinder.java
navinrathore: **before or after alignDupes.** How will the impact be different? after alignDupes, the work is simple and straightforward.

core/src/main/java/zingg/util/DSUtil.java
navinrathore: New class **zingg.preprocess.StopWords** has been introduced. That's what expected?

core/src/main/java/zingg/util/DSUtil.java
sonalgoyal: yes so why javardd - cant we do the same using dataset?

core/src/main/java/zingg/util/DSUtil.java
sonalgoyal: junit

core/src/main/java/zingg/util/DSUtil.java
```diff
@@ -253,4 +262,67 @@
 				.filter(f -> !(f.getMatchType() == null || f.getMatchType().equals(type)))
 				.collect(Collectors.toList());
 	}
+
+	public static Dataset<Row> preprocessForStopWords(SparkSession spark, Arguments args, Dataset<Row> ds) {
+
+		List<String> wordList = new ArrayList<String>();
+		Dataset<Row> modDS = ds;
```
sonalgoyal, line 269: just use ds as it is immutable anyways?

core/src/main/java/zingg/TrainingDataFinder.java
sonalgoyal: alignDupes is aligning the columns from the pairs. logically we should be working on the actual data here. also we can avoid the o_ approach which is more like a hack. let us do joins here. easier to understand

core/src/main/java/zingg/util/DSUtil.java
```diff
@@ -253,4 +262,67 @@
...
+	public static UserDefinedFunction removeStopWords(List<String> stopWords) {
+		return udf((String s) -> {
+			 		ArrayList<String> allWords = Stream.of(s.split(" "))
+						.collect(Collectors.toCollection(ArrayList<String>::new));
+						allWords.removeAll(stopWords);
+			return allWords.stream().collect(Collectors.joining(" "));
+			}, DataTypes.StringType);
+	}
+
+	public static Dataset<Row> postprocessForStopWords(Dataset<Row> actual, Dataset<Row> orig, Arguments args) {
```
sonalgoyal, line 295: let us select the z_ columns and join with orig. Much easier to understand

sonalgoyal (changes requested):

core/src/main/java/zingg/util/DSUtil.java
navinrathore: avoided javardd call

core/src/main/java/zingg/util/DSUtil.java
navinrathore: junit shall be added once, prima facie, code looks fine.

core/src/main/java/zingg/TrainingDataFinder.java
navinrathore: o_ approach has been dropped. selection of appropriate columns and then joining with the original is used.
It works fine after alignDupes().

core/src/main/java/zingg/TrainingDataFinder.java
```diff
@@ -131,13 +141,16 @@ public void execute() throws ZinggClientException {
...
 		dupesActual = Util.addUniqueCol(dupesActual, ColName.CLUSTER_COLUMN );		
 		Dataset<Row> dupes1 = DSUtil.alignDupes(dupesActual, args);
+		dupes1 = DSUtil.postprocessForStopWords(dupes1, sampleOrginal, args);
 		Dataset<Row> dupes2 = dupes1.orderBy(ColName.CLUSTER_COLUMN);
 		LOG.debug("uncertain output schema is " + dupes2.schema());
+		dupes2.show(8);
```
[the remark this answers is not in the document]
navinrathore: done

core/src/main/java/zingg/util/DSUtil.java
navinrathore: done

core/src/main/java/zingg/util/DSUtil.java
navinrathore: yes, that is only done. Important thing is at what step is done??

sonalgoyal:
please add respective junits so that we can close.

navinrathore:
New changes are: 
* For Linker, Original restoration changes.
* Fix for Documenter Junit
* Stop words specific Junit tests
* Stop Words documentation

sonalgoyal (changes requested):

core/src/main/java/zingg/preprocess/StopWords.java
```diff
@@ -0,0 +1,86 @@
...
+	protected static String name = "zingg.preprocess.StopWords";
+	public static final Log LOG = LogFactory.getLog(StopWords.class); 
+
+    public static Dataset<Row> preprocessForStopWords(SparkSession spark, Arguments args, Dataset<Row> ds) {
+
+		List<String> wordList = new ArrayList<String>();
+		for (FieldDefinition def : args.getFieldDefinition()) {
+			if (!(def.getStopWords() == null || def.getStopWords() == "")) {
+				Dataset<Row> stopWords = PipeUtil.read(spark, false, false, PipeUtil.getStopWordsPipe(args, def.getStopWords()));
+				wordList = stopWords.select("_c0").collectAsList().stream().map(f -> f.getString(0)).collect(Collectors.toList());
```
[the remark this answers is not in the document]
navinrathore: Modified the code

core/src/main/java/zingg/preprocess/StopWords.java
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	public static UserDefinedFunction removeStopWords(List<String> stopWords) {
+		return udf((String s) -> {
+			 		ArrayList<String> allWords = Stream.of(s.split(" "))
+						.collect(Collectors.toCollection(ArrayList<String>::new));
+						allWords.removeAll(stopWords);
+			return allWords.stream().collect(Collectors.joining(" "));
+			}, DataTypes.StringType);
+	}
```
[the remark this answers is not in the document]
navinrathore: Changed to pattern matching.

core/src/main/java/zingg/preprocess/StopWords.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public static Dataset<Row> postprocess(Dataset<Row> actual, Dataset<Row> orig) {
```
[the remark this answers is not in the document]
navinrathore: Moved.

core/src/main/java/zingg/preprocess/StopWords.java
```diff
@@ -0,0 +1,86 @@
...
+		cols.add(actual.col(ColName.CLUSTER_COLUMN));
+		cols.add(actual.col(ColName.ID_COL));
+		cols.add(actual.col(ColName.PREDICTION_COL));
+		cols.add(actual.col(ColName.SCORE_COL));
+		cols.add(col(ColName.MATCH_FLAG_COL));
+
+		Dataset<Row> zFieldsFromActual = actual.select(JavaConverters.asScalaIteratorConverter(cols.iterator()).asScala().toSeq());
+		
+		Dataset<Row> joined = zFieldsFromActual.join(orig, ColName.ID_COL);
+
```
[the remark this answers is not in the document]
navinrathore: COL_ID is already part of alignDupes() but not alignLinked(). so it has been introduced in postprocessLinked(), but deleted after join() step. I think ID is the only join-able column. Isn't it?

core/src/test/java/zingg/preprocess/TestStopWords.java
```diff
@@ -0,0 +1,245 @@
...
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.Metadata;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import zingg.client.Arguments;
```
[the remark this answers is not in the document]
navinrathore: Moved two methods.

core/src/test/java/zingg/preprocess/TestStopWords.java
```diff
@@ -0,0 +1,245 @@
...
+	@AfterClass
+	public static void teardown() {
+		if (ctx != null)
+			ctx.stop();
+		if (spark != null)
+			spark.stop();
+	}
+
+	@Test
+	public void testRemoveStopWords() {
```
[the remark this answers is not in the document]
navinrathore: Added @DisplayName, However, there could be an alternative @DisplayNameGeneraator based on the function name. Currently, only _ are replaced . CamelCase one is not there. but a extendable class is there that we can use. Or we can wait till future version
e.g. ifItIsDivisibleBy4ButNotBy100() -->>  if It Is Divisible By 4 But Not By 100

core/src/test/java/zingg/preprocess/TestStopWords.java
```diff
@@ -0,0 +1,245 @@
...
+							RowFactory.create("1648811730857:10", "10", "1.0", "0.555555", "-1",
+									"The zingg spark application", "two", "Yes. good application", "test"),
+							RowFactory.create("1648811730857:20", "20", "1.0", "1.0", "-1",
+									"It very popular data science", "Three", "true indeed", "test"),
+							RowFactory.create("1648811730857:30", "30", "1.0", "0.999995", "-1",
+									"It written java scala", "four", "", "test"),
+							RowFactory.create("1648811730857:40", "40", "1.0", "1.0", "-1", "Best luck zingg", "Five",
+									"thank", "test")),
+					schemaActual);
+
```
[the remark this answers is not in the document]
navinrathore: There are two test cases already that covers the same intention. However, more test cases would be written after discussion.
LowerCases testing of stop words and the text has been covered in test cases.

core/src/test/java/zingg/preprocess/TestStopWords.java
```diff
@@ -0,0 +1,245 @@
...
+	@AfterClass
+	public static void teardown() {
+		if (ctx != null)
+			ctx.stop();
+		if (spark != null)
+			spark.stop();
+	}
+
+	@Test
```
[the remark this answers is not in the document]
navinrathore: Added @DisplayName, However, there could be an alternative @DisplayNameGeneraator based on the function name. Currently, only _ are replaced . CamelCase one is not there. but a extendable class is there that we can use. Or we can wait till future version
e.g. ifItIsDivisibleBy4ButNotBy100() -->> if It Is Divisible By 4 But Not By 100

navinrathore:
Needless to say, the code need to be refactored when other preprocessing requirements are added. Similarly, Test cases should also be elaborated.

navinrathore:
In html template, set width of columns to auto (to fit content length), not to page width as earlier
In csv template, count of stopword has been added
StopWords.java, if cell is null, do nothing. It was an issue
Documentor.java, 
	a) blank html files for the z columns to have conssitency in generated html file for all columns
	b) moved the stopword html/csv directories verification and creation out of every call of findAndWriteStopWords.

sonalgoyal (commented):
- test for Mobile/T-Mobile pattern
- stop words should have header - StopWord, Count
- read from StopWord column not _c0
- https://stackoverflow.com/questions/40038575/how-to-match-the-forward-slash-using-regex

PR #209, opened 2022-04-15
New InMemoryPipe has been added

navinrathore:
The new pipe can take inputs from memory and write the output in the memory.

### DIFF
--- a/api/scala/FebrlExample.scala
+++ b/api/scala/FebrlExample.scala
@@ -1,0 +1,110 @@
+import zingg.client._;
+import java.util.ArrayList;
+
+//setting silent mode for Argument creation only 
+:silent
+val args = new Arguments();
+
+val fname = new FieldDefinition();
+fname.setFieldName("fname");
+fname.setDataType("\"string\"");
+fname.setMatchType(MatchType.FUZZY);
+fname.setFields("fname");
+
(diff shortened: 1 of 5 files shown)


api/scala/FebrlExample.scala
```diff
@@ -0,0 +1,110 @@
...
+args.setLabelDataSampleSize(0.5f);
+
+//reading dataset into inputPipe and settint it up in 'args'
+//below line should not be required
+val df = spark.read.option("header", "true").csv("examples/febrl/test.csv")
+import zingg.client.pipe.InMemoryPipe;
+import java.util.HashMap
+
+val inputPipe = new InMemoryPipe(df);
+inputPipe.setProps(new HashMap[String, String]());
```
sonalgoyal, line 94: why do we need this?

sonalgoyal (changes requested):

api/scala/FebrlExample.scala
navinrathore: It doesn't work without it. crashes at (Line)[https://github.com/zinggAI/zingg/blob/streamsets/core/src/main/java/zingg/util/PipeUtil.java#L64]
In output pipe this doesn't needed.

Issue #218, opened 2022-04-22
--showConcide=true or --showConcise true is not working on main


sonalgoyal:
please verify @navinrathore

navinrathore:
Reason is MatchType.java contains enum :-
 DONT_USE("DONT USE"); 

But we normally use "DONT_USE"  [with underscore in between] in config file. It should be made consistent.

navinrathore:
DONT_USE should be used everywhere.

PR #222, opened 2022-04-25
MatchType 'DONOT USE' updated in the docs and TC added

navinrathore:
Updated docs with correct usage of "DONT USE"
Also added TC for DSUtil::getFieldDefColumns()


navinrathore:
#218

PR #233, opened 2022-05-03
Added new format 'bigquery'

navinrathore:


navinrathore:
Source data configuration may look like below
```
		"data" : [{
			"name":"test", 
			"format":"bigquery", 
			"props": {
				"credentialsFile": "/home/work/product/final/zingg-1/mynotification-46566-905cbfd2723f.json",
				"table": "mynotification-46566.zinggdataset.zinggtest",
				"viewsEnabled": true			
			}
		}],
```

PR #244, opened 2022-05-10
Documenter JUnit tests

navinrathore:
Separate test files for Model, Column and (Main) Documenter classes.
use datasets/model  iTunes-amazon (model id 105), amazon-google (103), febrl (100)

### DIFF
--- a/core/src/test/java/zingg/documenter/TestColumnDocumenter.java
+++ b/core/src/test/java/zingg/documenter/TestColumnDocumenter.java
@@ -1,0 +1,58 @@
+package zingg.documenter;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import java.io.File;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import zingg.client.Arguments;
+import zingg.client.MatchType;
+import zingg.client.ZinggClientException;
+import zingg.util.DSUtil;
+
+public class TestColumnDocumenter extends TestBaseDocumenter {
+  
+	public static final Log LOG = LogFactory.getLog(TestColumnDocumenter.class);
+
+	@BeforeEach
+	public void setUp() {
+
+		try {
+			args = Arguments.createArgumentsFromJSON(getClass().getResource("/testConfig.json").getFile());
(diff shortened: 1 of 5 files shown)


core/src/test/java/zingg/documenter/TestColumnDocumenter.java
```diff
@@ -0,0 +1,58 @@
...
+		} catch (ZinggClientException e) {
+			e.printStackTrace();
+		}
+
+		File directory = new File(args.getZinggDocDir());
+		LOG.info("Directory Name: " + args.getZinggDocDir());
+		assertEquals(true, directory.exists(), "Zingg Directory does not exist. Docs haven't been generated. " + args.getZinggDocDir());
+
+		File[] listOfFiles = directory.listFiles((dir, name) -> name.endsWith(".html"));
+		int htmlDocsCount = DSUtil.getFieldDefinitionFiltered(args, MatchType.DONT_USE).size()
```
sonalgoyal, line 48: dont call other class methods in junits

core/src/test/java/zingg/documenter/TestModelDocumenter.java
```diff
@@ -0,0 +1,47 @@
...
+			e.printStackTrace();
+			System.out.println("Unexpected exception received " + e.getMessage());
+			fail(e.getMessage());
+		}
+
+		ModelDocumenter md = new ModelDocumenter(spark, args);
+		try {
+			md.process();
+		} catch (ZinggClientException e) {
+			e.printStackTrace();
```
sonalgoyal, line 37: wont you fail here? also see if both try catch can be combined?

core/src/test/java/zingg/documenter/TestModelDocumenter.java
```diff
@@ -0,0 +1,47 @@
...
+		}
+
+		File modelFile = new File(args.getZinggDocFile());
+		assertEquals(true, modelFile.exists(), "Doc file hasn't been generated. " + args.getZinggDocFile());
+		assertModelFleSize(modelFile.length() > DEFAULT_FILE_SIZE, modelFile.length());
+
+		File directory = new File(args.getZinggDocDir());
+		assertEquals(true, directory.exists(), "Zingg Directory does not exist. Column docs haven't been generated. " + args.getZinggDocDir());
```
sonalgoyal, line 45: can we add some tests to verify structure and content of the docs generated? you may need to modify the code to have unit methods which can be tested

sonalgoyal (changes requested):
added comments, please check

Issue #237, opened 2022-05-05
Reduce Zingg jar size


sonalgoyal:
now that we have a way to specify the external jars, we should remove snowflake, mysql and other dependent jars which are being rolled in the shaded jar.

sonalgoyal:
current jar size is way to much - let us remove all the non necessary dependencies.

PR #248, opened 2022-05-11
Removed dependencies of snowflake, mysql, cassandra, elastic, apache-poi, log4j from pom

navinrathore:
Dependencies of snowflake, mysql etc. have been removed from pom. However, their support in Zingg continues to be available. The environment varible **ZINGG_EXTRA_JARS** must be set to point to the jars of concerned data source. More in  Zingg [docs](https://docs.zingg.ai/zingg/connectors/snowflake)


navinrathore:
#237

Issue #256, opened 2022-05-12
Z columns doc should be generated from different template 


navinrathore:
**Description**
* Different template should be used to generate document for 'z' columns
* For 'z' columns, stop words should be generated
* As of now, doc for a z columns can have a simple text like "The field <z_column> is internally used by Zingg." The message should be supplied by the application.

sonalgoyal:
EDIT: stop words should NOT be generated for z columns

navinrathore:
Yes. did that

sonalgoyal:
after discussion with @navinrathore , the plan is to go with the links and add descriptions of each column

z_cluster - identifies a group of records which match or dont match with each other. For each group, z_cluster is unique. Member records of a group share the same z_cluster. 

z_prediction - Wjhat Zingg thinks about this group/pair of records - 0 for not a match, 1 for a match

z_score - the probability of a pair of records matching. The higher the score, the more likely they are a match.

z_isMatch - this is the label provided by the user

navinrathore:
z_id - an internal id given by Zingg to uniquely identify the record
z_source - the source of data as set in the name property of the data in the Zingg configuration file

navinrathore:
#257

PR #257, opened 2022-05-13
Z columns doc to have different template.

navinrathore:
Zingg internal columns (Z columns) should have different template.


core/src/main/java/zingg/documenter/ColumnDocumenter.java
```diff
@@ -86,4 +92,24 @@ public Map<String, Object> addStopWords(Dataset<Row> data, String fieldName, Map
 		
 		return params;
 	}
+
+	public String getColumnBaseContent (String col) {
+		String message = "The field '" + col + "' is internally used by Zingg.";
+
+		if (col.equals(ColName.CLUSTER_COLUMN)) {
```
sonalgoyal, line 99: This needs to go in the template, not in the java code

core/src/main/java/zingg/documenter/ColumnDocumenter.java
```diff
@@ -26,6 +26,7 @@ public class ColumnDocumenter extends DocumenterBase {
 
 	private final String CSV_TEMPLATE = "stopWordsCSV.ftlh";
```
sonalgoyal, line 27: One thing I seem to have missed in my earlier reviews - we should have column templates which include stop word templates. Then it much easier to keep stuff modular and extend etc as needed

sonalgoyal (changes requested):

core/src/main/java/zingg/documenter/ColumnDocumenter.java
navinrathore: moved content in a separate template and included in the parent template

core/src/main/java/zingg/documenter/ColumnDocumenter.java
navinrathore: Moved to Template in if else blocks.

navinrathore:
#256

sonalgoyal (changes requested):
a. z col list should just be columns that star with z_
b. StopWordDocumenter should be introduced and included in the ColumnDocumenter. All code and logic for StopWords should go in there. We can then add more documenters for columns. eg data quality. 
c.

navinrathore:
> a. z col list should just be columns that star with z_ b. StopWordDocumenter should be introduced and included in the ColumnDocumenter. All code and logic for StopWords should go in there. We can then add more documents for columns. eg data quality.

b. Moved the StopWords stuff in a separate class
a.  z col list is indeed formed from 6 z_ columns. Is there any issue in it?

sonalgoyal:
It will be neater to have a check in the pattern for z columns instead of keeping a list

sonalgoyal:
Please also rebase this pr as there is a merge conflict

navinrathore:
> It will be neater to have a check in the pattern for z columns instead of keeping a list

We go for documentation with marked data as input.  It includes field definition fields and only few z_columns. DONT_USE fields also excluded.
But, Documenter is generating docs for all fields (excludes DON'T USE) but includes all z_fields. As of now, only ones part of marked data are shown with hyperlink heading. 
There was some issue working with sequence of columns of marked data. then we moved to get information from FieldDefinition and z Col list.

navinrathore:
> Please also rebase this pr as there is a merge conflict

resolved conflict.

Issue #263, opened 2022-05-17
generateDocs should be independent of the data


sonalgoyal:
Our current logic of generating documents breaks when we only have the model, and not the data. This happens often, as we ask customers to ship their model folder to us for debugging. In this case, we do not have the data at all.

sonalgoyal:
Conceptually, if there is no data, do nothing for data. If there is data, we should generate stop words here. So Documenter, ModelDocumenter and Data Documenter. Data Documenter will invoke the StopWordsDocumenter.

sonalgoyal:
ColumnDocumenter to be renamed as ModelColDocumenter

sonalgoyal:
Have another Documenter ColumnDocumenter which lists field type. We will use it later to profile data etc. This should invoke the stopworddocumenter.

PR #270, opened 2022-05-17
GenerateDocs becomes independent of Data

navinrathore:
Separated Model Documenter and Data Documenter
StopWords Becomes part of Data Documenter
Got rid of fixed list of 'z' columns

Refer to bug #263


navinrathore:
Largely, all things have been incorporated and tested, that came in our discussion. Still, following points need attention.
* Column/Data Document - parent doc Link (parent Name) is hard coded (Top right corner on page). Need to know root location ($ZINGG_HOME)
* Decision on Column documents generated from modelDocumenter and DataDocumenter differentiation and their location of storage. As of now, both have same name and location. first ones are overwritten  by stop words one. Original ones do not have any information yet. Though, all links work fine.
* Data documentation - list of schema field formation, table beautification
* Column/Data documentation - additional data on the page, if any

Issue #279, opened 2022-05-20
secondstring dependency is not getting resolved


navinrathore:
**Describe the bug**
secondstring dependency is not getting installed with zingg pom for a specific environment. 

Following error is thrown
java.lang.ClassNotFoundException: com.wcohen.ss.AbstractStringDistance

secondstring  is not getting added into Zingg Jar.

**Workaround**
It has been installed by 
```
mvn org.apache.maven.plugins:maven-install-plugin:3.0.0-M1:install-file -Dfile=thirdParty/lib/secondstring.jar
                      -DgroupId=[com.wcohen.ss](http://com.wcohen.ss/)                                                                                  -DartifactId=secondstring
                          -Dversion=2021                                                                                  -Dpackaging=jar
           -DlocalRepositoryPath=/home/raviraj/.m2/repository
```

navinrathore:
Maven has been reinstalled through  **sudo apt install maven**
The version 3.6.3 got installed. Thereafter, things have been working as expected. no workaround needed anymore.
```
INFO] --- maven-install-plugin:2.5.2:install-file (install-external) @ parent ---
[INFO] pom.xml not found in secondstring.jar
[INFO] Installing /home/raviraj/zingg/thirdParty/lib/secondstring.jar to /home/raviraj/.m2/repository/com/wcohen/ss/secondstring/2021/secondstring-2021.jar
[INFO] Installing /tmp/mvninstall8972178536806859739.pom to /home/raviraj/.m2/repository/com/wcohen/ss/secondstring/2021/secondstring-2021.pom
```

sonalgoyal:
is the dev env doc updated for the right version of maven?

sonalgoyal:
seems we made some errors while configuring 3.3.9.

navinrathore:
The below method is good to install maven. It shall install latest version
```
sudo apt install maven
```
#282

PR #282, opened 2022-05-23
Updated installation document to install maven using sudo apt

navinrathore:
Changed to standard installation method for maven that is using "apt install".
Any version of maven shall work. Latest one should be preferred.(on date, it is 3.6.3)


navinrathore:
#279

PR #290, opened 2022-05-25
Match type pin code

RavirajBaraiya:


core/src/main/java/zingg/similarity/function/PinCodeMatchTypeFunction.java
```diff
@@ -0,0 +1,58 @@
...
+	public PinCodeMatchTypeFunction() {
+		this("PinCodeMatchTypeFunction");		
+	}
+	
+	public PinCodeMatchTypeFunction(String s) {
+		super(s);
+	}
+	
+	@Override
+	public Double call(String first, String second) {
```
sonalgoyal, line 23: just split the string, take the first parts and call super. that will invoke the stringsimilarity's call which has the code to match two stirngs.

core/src/test/java/zingg/similarity/function/TestPinCodeMatchTypeFunction.java
```diff
@@ -0,0 +1,96 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TestPinCodeMatchTypeFunction {
+	
+	
+	@Test
+	public void testFirstPinNull() {
+		PinCodeMatchTypeFunction PinMatch = new PinCodeMatchTypeFunction();
```
sonalgoyal, line 15: variable naming with small case. please rename to pincodeMatchFn

sonalgoyal: apply to all

core/src/test/java/zingg/similarity/function/TestPinCodeMatchTypeFunction.java
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	@Test
+	public void testBothNull() {
+		PinCodeMatchTypeFunction PinMatch = new PinCodeMatchTypeFunction();
+		assertEquals(1d, PinMatch.call(null, null));
+	}
+
+	@Test
+	public void testBothNotEmptysame1() {
```
sonalgoyal, line 49: rename to testBothNotEmptyExact - s cap

core/src/test/java/zingg/similarity/function/TestPinCodeMatchTypeFunction.java
```diff
@@ -0,0 +1,96 @@
...
+		PinCodeMatchTypeFunction PinMatch = new PinCodeMatchTypeFunction();
+		assertEquals(1d, PinMatch.call("45678", "45678"));
+	}
+
+	@Test
+	public void testBothNotEmptydiff1() {
```
sonalgoyal, line 55: camel casing - see above comment for name of method

sonalgoyal: testbothNotEmptyDifferent

core/src/test/java/zingg/similarity/function/TestPinCodeMatchTypeFunction.java
```diff
@@ -0,0 +1,96 @@
...
+		PinCodeMatchTypeFunction PinMatch = new PinCodeMatchTypeFunction();
+		assertEquals(0d, PinMatch.call("34234", "34334"));
+	}
+
```
sonalgoyal, line 59: rename all other test cases accordingly - first part match, both parts match, second part doesnt match etc...

sonalgoyal (changes requested):
Very impressive first contribution @RavirajBaraiya ! Adding small review comments, please update and send pr. thanks!

Issue #141, opened 2022-02-11
email match type seems to be missing


sonalgoyal:
couldnt find corresponding feature def in the code. need to check

sonalgoyal:
In our config, we define different ways in which we can match an attribute - fuzzy, exact etc. we seem to have defined an email type which was supposed to look at only the first part of the email and ignore the domain. But I do not see it’s implementation in the code

abhioncbr:
Please assign the issue to me. I will give it a try.

sonalgoyal:
fixed through #291

PR #291, opened 2022-05-25
Match type email

RavirajBaraiya:
issue #141

### DIFF
--- a/core/src/main/java/zingg/similarity/function/EmailMatchTypeFunction.java
+++ b/core/src/main/java/zingg/similarity/function/EmailMatchTypeFunction.java
@@ -2,13 +2,14 @@ package zingg.similarity.function;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.spark.ml.util.Identifiable$;
 
 
-
-public class EmailMatchTypeFunction extends StringSimilarityFunction {
+public class EmailMatchTypeFunction extends StringSimilarityDistanceFunction {
 
 	public static final Log LOG = LogFactory
 			.getLog(EmailMatchTypeFunction.class);
+	private static final String String = null;
 	
 	
 	public EmailMatchTypeFunction() {
@@ -17,10 +18,12 @@ public class EmailMatchTypeFunction extends StringSimilarityFunction {
 	
 	public EmailMatchTypeFunction(String s) {
 		super(s);
+		//gap = new SAffineGap();
 	}
 	
 	//@Override
-	public  EmailMatchTypeFunction(String first, String second) {
+	public EmailMatchTypeFunction(String first, String second) {
+		super(String,String);
 		first = first.split("@",0)[0];
 		second = second.split("@",0)[0];
 		super.call(first,second);
--- a/core/src/main/java/zingg/similarity/function/EmailMatchTypeFunction.java
+++ b/core/src/main/java/zingg/similarity/function/EmailMatchTypeFunction.java
@@ -1,0 +1,59 @@
+package zingg.similarity.function;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+
+
+public class EmailMatchTypeFunction extends StringSimilarityFunction {
+
+	public static final Log LOG = LogFactory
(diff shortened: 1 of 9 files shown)


core/src/main/java/zingg/similarity/function/EmailMatchTypeFunction.java
```diff
@@ -0,0 +1,59 @@
+package zingg.similarity.function;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+
+
+public class EmailMatchTypeFunction extends StringSimilarityFunction {
```
sonalgoyal, line 8: extend StringSimilarityDistanceFunction and use that

core/src/test/java/zingg/similarity/function/TestEmailMatchTypeFunction.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Arrays;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TestEmailMatchTypeFunction {
+	
+	
+	@Test
```
sonalgoyal, line 13: please update test scores asserts accordingly. you can invoke the StringSimilarityDistanceFunction with the first part of email and check that you are getting what it returns

sonalgoyal (changes requested):

core/src/test/java/zingg/similarity/function/TestEmailMatchTypeFunction.java
RavirajBaraiya: I tried to do that. but, It is giving me this error: Cannot instantiate the type StringSimilarityDistanceFunction
should I leave that part by checking scores with EmailMatchTypeFunction ?

sonalgoyal: ok, sorry I realized later it is an abstract class. Can you please extend AffineGapSimilarityFunction and use that in the tests as well?

RavirajBaraiya: Ok, let me try that.

core/src/test/java/zingg/similarity/function/TestEmailMatchTypeFunction.java
```diff
@@ -12,61 +12,101 @@ public class TestEmailMatchTypeFunction {
...
 	@Test
 	public void testFirstEntryNull() {
-		EmailMatchTypeFunction emailMatchFn = new EmailMatchTypeFunction();
-		assertEquals(1d, emailMatchFn.call(null, "xyz321@pqr.co"));
+		AffineGapSimilarityFunction emailMatchFn = new AffineGapSimilarityFunction();
+		String first = null;
+		String second = "xyz321@pqr.co";
+		//first = first.split("@",0)[0];
+		second = second.split("@",0)[0];
+		assertEquals(1d, emailMatchFn.call(first, second));
```
sonalgoyal, line 20: you need to verify here that the output from emailmatchfn is the same as the output from first part of emails sent to affinegap fn

core/src/test/java/zingg/similarity/function/TestEmailMatchTypeFunction.java
```diff
@@ -12,61 +12,101 @@ public class TestEmailMatchTypeFunction {
...
 
 	@Test
 	public void testFirstEntryEmpty() {
 		EmailMatchTypeFunction emailMatchFn = new EmailMatchTypeFunction();
-		assertEquals(1d, emailMatchFn.call("", "xyz321@pqr.co"));
+		String first = "";
+		String second = "xyz321@pqr.co";
+		first = first.split("@",0)[0];
+		second = second.split("@",0)[0];
+		assertEquals(1d, emailMatchFn.call(first, second));
```
sonalgoyal, line 30: the function does the split here, you should not be splitting unless you want to send the split values to AffineGapSimilarityFunction 

so validate that original first and second when passed to emailfn returns the same as first part of split string sent to affinegap similarity, ok?

sonalgoyal (changes requested):
test cases need a relook.

Issue #261, opened 2022-05-16
more blocking functions for integer and double types


sonalgoyal:
We need to add more blocking functions for Integer and Double types other than the double and exact ones that we have

In all cases, handle 0 and Nan and negative numbers appropriately.
Both
- number without last digit. eg for integer type 234 and 233 will return 23, double type - 234.5 and 233.5 will return 23
- range functions. return 0 if num < 10, 1 if 10 < num < 100, 2 if less than 1000 but greater than 100, 3 if less than 10000(this can be odelled as FirstChars/LastChars functions, one base class and one derived class for each range)
- lessThan 0 function

extra functions for Double type
- number with one digit after decimal (23.4445 gives 23.44)
- number with 2 digits after decimal
- 3 digits after decimal
(both above to be modeled with one base class like firstchars)

sonalgoyal:
please have appropriate junits for each - check the similarity function junits for an idea of how to build these junits.

navinrathore:
Code changes made. Junits added.

sonalgoyal:
i dont see the pr?

PR #292, opened 2022-05-25
More blocking functions

navinrathore:
#261


navinrathore:
Tested findTrainingData, Train etc. with 1 field for integer, double.
separate commits  for different type of blocking functions. They can be views commit wise.

core/src/main/java/zingg/hash/TruncateDouble.java
```diff
@@ -0,0 +1,31 @@
+package zingg.hash;
+
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.api.java.UDF1;
+import org.apache.spark.sql.types.DataTypes;
+
+public class TruncateDouble extends HashFunction implements UDF1<Double, Double> {
+	int numDecimalPlaces;
+	static final int[] POWERS_OF_10 = {1, 10, 100, 1000, 10000, 100000};
```
sonalgoyal, line 9: why use this? just say 10 exp the argument.

core/src/test/java/zingg/hash/TestRangeDbl.java
```diff
@@ -0,0 +1,95 @@
...
+		value = new RangeBetween1000And10000Dbl();
+		assertEquals(0, value.call((double) 65));
+	}
+
+	@Test
+	public void testRangeForValue867() {
+		RangeDbl value = new RangeBetween0And10Dbl();
+		assertEquals(0, value.call((double) 867));
+		value = new RangeBetween10And100Dbl();
+		assertEquals(0, value.call((double) 867));
```
sonalgoyal, line 76: saying 867d makes the code more readable..upto you if you feel that way.

core/src/test/java/zingg/hash/TestRangeDbl.java
```diff
@@ -0,0 +1,95 @@
+package zingg.hash;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
sonalgoyal, line 5: tests for different classes should be in different junit classes. multiple asserts should not go into same unit test, as you dont know then which one broke. If you think it is too much to write junits for the derived classes, just test the main base class.

core/src/test/java/zingg/hash/TestRangeInt.java
```diff
@@ -0,0 +1,93 @@
+package zingg.hash;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TestRangeInt {
+
+	@Test
+	public void testRangeForValueZero() {
```
sonalgoyal, line 10: same comment as earlier

sonalgoyal (changes requested):

core/src/main/java/zingg/hash/TruncateDouble.java
navinrathore: pow() seems a relatively expansive operation.
https://stackoverflow.com/questions/46983772/fastest-way-to-obtain-a-power-of-10

core/src/test/java/zingg/hash/TestRangeDbl.java
navinrathore: Shall Make the changes

core/src/test/java/zingg/hash/TestRangeDbl.java
navinrathore: I'll change

core/src/test/java/zingg/hash/TestRangeInt.java
navinrathore: I'll make change

Issue #269, opened 2022-05-17
move getRecords and setRecords inn InmmemmoryyPipe to Pipe class


sonalgoyal:
current pipes have no notion of records they represent, making testing hard

sonalgoyal:
there are two methods and one attribute getRecords and setRecords in InMemoryPipe class. We should move them to the base class Pipe. 
Then we should test the scala api at https://github.com/zinggAI/zingg/tree/main/api/scala. To do that, open spark-shell with --jars option 
spark-shell --jars <path to zingg jar>

Run the scala code above and see that you get an output.

PR #293, opened 2022-05-25
resolved errors

chetan453:
issue #269


api/scala/FebrlExample.scala
```diff
@@ -110,6 +109,9 @@ val options = new ClientOptions("--phase", "match",  "--conf", "dummy", "--licen
 //Zingg execution for the given phase
 val client = new Client(args, options);
 client.init();
+System.out.println("line 113");
```
sonalgoyal, line 112: these printlns should not be in the code

Issue #260, opened 2022-05-16
Revisit zingg.sh


navinrathore:
**Description**
"zing.conf" file should include other zingg variables as well. It should be central point for defining zingg configuration.
zingg.sh should be changed accordingly

Ref: PR #253, issue #15

sonalgoyal:
@navinrathore please check https://github.com/zinggAI/zingg/issues/243#issuecomment-1140209933

navinrathore:
As of now, spark configuration can be read from the zingg.conf file. Through the same file, environment variables could also be defined.

PR #301, opened 2022-05-30
support for python phases in zingg.sh

navinrathore:
Tested with cluster as well.
If phase is not on the list of dictionary (hard coded), standard Zingg execution take place. In case of python phase, it shall be executed. only phase option is used as of now.

### DIFF
--- a/scripts/zingg.sh
+++ b/scripts/zingg.sh
@@ -51,7 +51,9 @@ while [[ $# -gt 0 ]]; do
 			PHASE="$2"
 			is_python_phase $PHASE # check if it is a python phase
 			RUN_PYTHON_PHASE=$? # store the return status
-			PYTHON_PHASE_EXE=${PYTHON_PHASES[$PHASE]}
+			if [[ $RUN_PYTHON_PHASE -eq 0 ]]; then
+				PYTHON_PHASE_EXE=${PYTHON_PHASES[$PHASE]}
+			fi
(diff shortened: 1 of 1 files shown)


scripts/zingg.sh
```diff
@@ -3,34 +3,77 @@
...
+	esac
+done
+set -- "${POSITIONAL_ARGS[@]}" # restore positional parameters
+
 OPTION_SPARK_CONF+=$(read_zingg_conf)
+
+# if it is a python phase
+if [[ $RUN_PYTHON_PHASE -eq 0 ]]; then
+	echo "Executing python phase: ${PHASE} (${PYTHON_PHASE_EXE})"
+	$SPARK_HOME/bin/spark-submit --master $SPARK_MASTER --jars=$ZINGG_JARS --conf spark.driver.extraClassPath=$ZINGG_JARS --conf spark.executor.extraClassPath=$ZINGG_JARS $PYTHON_PHASE_EXE --email $EMAIL --license $LICENSE
```
sonalgoyal, line 73: we need to pass the remaining arguments to the python phase as well.

sonalgoyal: how about passing the other options? (ALL_OPTIONS)

sonalgoyal: and --py-files?

sonalgoyal (changes requested):

navinrathore:
#260

PR #303, opened 2022-05-30
env variables can be defined in zingg.conf in addition to spark properties

navinrathore:
additional work in zingg.sh


navinrathore:
#260

Issue #262, opened 2022-05-17
Technical Debt


sonalgoyal:
*LabelUpdater and Updater - exit value(9)is hardcoded.  Should be declared a String at the top and used everywhere.

dipan121003:
Hi, I'd like to work on this issue. Can you assign it to me?

PR #304, opened 2022-05-31
ISSUE-262 Clean code in updater and label updater

ranjithmuddana:
Code Changes for #262

### DIFF
--- a/core/src/main/java/zingg/Labeller.java
+++ b/core/src/main/java/zingg/Labeller.java
@@ -22,6 +22,8 @@ import zingg.util.LabelMatchType;
 public class Labeller extends ZinggBase {
 
 	protected static String name = "zingg.Labeller";
+	protected static final int INT_NINE = 9;
+	protected static final String NINE = String.valueOf(INT_NINE);
 	public static final Log LOG = LogFactory.getLog(Labeller.class);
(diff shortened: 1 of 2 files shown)


core/src/main/java/zingg/Labeller.java
```diff
@@ -160,9 +162,9 @@ int readCliInput() {
 		System.out.println("\tYes, they match       : 1");
 		System.out.println("\tNot sure              : 2");
 		System.out.println();
-		System.out.println("\tTo exit               : 9");
+		System.out.println("\tTo exit               : " + NINE);
 		System.out.println();
-		System.out.print("\tPlease enter your choice [0,1,2 or 9]: ");
+		System.out.print("\tPlease enter your choice [0,1,2 or " + NINE + "]: ");
 
 		while (!sc.hasNext("[0129]")) {
```
sonalgoyal, line 169: Thank you for your code contribution @ranjithmuddana ! Here are a few suggestions

- though the issue talked about 9, we have a general problem with 0, 1, 2 and 9. 0, 1 and 2 should come from the ColValues class. Maybe we can add the constant EXIT there as well with value 9. So we dont have to hard code any of the values, like in line 162, 163, , 167, 169. 
Does that make sense?

sonalgoyal (changes requested):

PR #307, opened 2022-05-31
throw error if format is null; happens due to invalid/unsopprted format configuration

navinrathore:
An alternative is if we can throw error at the time of parsing config file when we find that format is not included in zingg list.
Also, we may remove restriction of support of only fixed list of formats. Let Spark throw error if it is an invalid/unsupported format.

Testcases for valid as well as invalid format added.

### DIFF
--- a/core/src/test/java/zingg/util/TestPipeUtil.java
+++ b/core/src/test/java/zingg/util/TestPipeUtil.java
@@ -1,0 +1,55 @@
+package zingg.util;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.spark.sql.SparkSession;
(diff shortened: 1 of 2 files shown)


core/src/test/java/zingg/util/TestPipeUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import zingg.ZinggSparkTester;
+import zingg.client.pipe.FilePipe;
+import zingg.client.pipe.Format;
+import zingg.client.pipe.Pipe;
+
+public class TestPipeUtil extends ZinggSparkTester {
+	public static final Log LOG = LogFactory.getLog(TestPipeUtil.class);
+
+	@Test
+	public void testGetDataFRameReaderforGivenFormat() throws Exception {
```
sonalgoyal, line 23: why are we testing with reflection? why not invoke the method directly?

sonalgoyal: same comment for second testcase

sonalgoyal (changes requested):
comments added.

sonalgoyal (changes requested):
see comments

navinrathore:
* The PR #370 supersede this PR. Hence, this is being closed.

Issue #306, opened 2022-05-31
figure out running user scripts from zingg.sh


sonalgoyal:
say we have a user who uses the zingg package to build the config and run the code. also ensure spark master is getting passed.

PR #311, opened 2022-06-06
Python User script support in zingg.sh

navinrathore:
figure out running user scripts from zingg.sh #306

### DIFF
--- a/scripts/zingg.sh
+++ b/scripts/zingg.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-#ZINGG_HOME=./assembly/target
+ZINGG_HOME=./assembly/target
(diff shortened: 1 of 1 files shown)


navinrathore:
Please add if additional/other functionality needed.

scripts/zingg.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-#ZINGG_HOME=./assembly/target
+ZINGG_HOME=./assembly/target
```
sonalgoyal, line 2: why do we need this?

sonalgoyal (changes requested):

scripts/zingg.sh
navinrathore: Reverted. erroneously, it was removed.

Issue #296, opened 2022-05-26
New phase assessModel


sonalgoyal:
Write a python script which whill expose the model stats - confusion matrix and number of records marked, unmarked, matches, non matches, not sure. 

We will use the Labeller class. The python script takes the conf and passes it to the Client. Client will invoke the Labeller. Refer to the python api example at https://github.com/zinggAI/zingg/blob/main/api/scala/FebrlExample.py.

The script calls getMarkedRecords, getMarkedRecordsStat, getUnmarkedRecords on the Client and provides the stats. You can convert the df returned by  the Client to python df. To build the confusion matrix, following can be used. 

import pandas as pd
import seaborn as sn
import matplotlib.pyplot as plt
 
confusion_matrix = pd.crosstab(markedRecords['z_isMatch'], markedRecords['z_prediction'], rownames=['Actual'], colnames=['Predicted'])
 
sn.heatmap(confusion_matrix, annot=True)
plt.show()

sonalgoyal:
I have added new methods on the client - getMatchedMarkedRecordsStat(Dataset<Row> markedRecords), getUnmatchedMarkedRecordsStat(Dataset<Row> markedRecords), getUnsureMarkedRecordsStat and getMarkedRecords()

you can use them to build the logic

sonalgoyal:
@RavirajBaraiya

navinrathore:
Confusion Matrix looks  like below
![image](https://user-images.githubusercontent.com/19220734/172234688-293e9f52-6988-4518-8bca-20eace4cd394.png)

navinrathore:
Generated Config File from Arguments object
[ArgumentsToFile.txt](https://github.com/zinggAI/zingg/files/8847335/ArgumentsToFile.txt)

navinrathore:
Statistics for model 100
```
No. of Records Marked   :  76
No. of Records UnMarked :  72
No. of Matches          :  14
No. of Non-Matches      :  24
No. of Not Sure         :  0
```

sonalgoyal:
need to look at the right model internally for this - should be expose label model or should we expose the actual model

PR #313, opened 2022-06-06
Updates in Python classes and 'assessModel' python phase

navinrathore:
Implemented requirements listed in 
support preparation New (https://github.com/zinggAI/zingg/discussions/289)  and
New phase assessModel (https://github.com/zinggAI/zingg/issues/296)


navinrathore:
* PeekModel implementation is due.
* Testcase to write config file based on Arguments object [incomplete due to a serialization issue]

core/src/main/java/zingg/ZinggBase.java
```diff
@@ -34,49 +34,49 @@
 
 public abstract class ZinggBase implements Serializable, IZingg {
```
sonalgoyal, line 35: please change ide settings so that tabs/spaces are consistent with older code and do not show up in the diff

python/phases/assessModel.py
```diff
@@ -6,10 +6,10 @@
 from IPython.display import display
```
sonalgoyal, line 7: this py script has to be invoked as zingg.sh --phase assessModel --conf conf.json. Arguments can not coded here as then user can not use it for their data.

python/phases/assessModel.py
```diff
@@ -18,20 +18,33 @@
...
+client = Zingg(args, options)
 client.init()
 client.execute()
+
 jMarkedDF = client.getMarkedRecords()
-print(jMarkedDF)
-markedDF = DataFrame(jMarkedDF, sqlContext)
-print(markedDF)
-pMarkedDF = markedDF.toPandas()
+markedDF = client.getDfFromDs(jMarkedDF)
```
sonalgoyal, line 28: why do you need markedDF? same for unmarked

python/phases/assessModel.py
```diff
@@ -18,20 +18,33 @@
...
+unmatched_marked = client.getUnmatchedMarkedRecordsStat(jMarkedDF)
+unsure_marked = client.getUnsureMarkedRecordsStat(jMarkedDF)
+print()
+print("No. of Records Marked   : ", total_marked)
+print("No. of Records UnMarked : ", total_unmarked)
+print("No. of Matches          : ", matched_marked)
+print("No. of Non-Matches      : ", unmatched_marked)
+print("No. of Not Sure         : ", unsure_marked)
 
+confusion_matrix = pd.crosstab(pMarkedDF['z_isMatch'], pMarkedDF['z_prediction'], rownames=['Actual'], colnames=['Predicted'])
```
sonalgoyal, line 48: please refer to java constants of col names here.

python/phases/zingg.py
```diff
@@ -32,47 +56,76 @@ def setFieldDefinition(self, fieldDef):
...
+
     def setLabelDataSampleSize(self, labelDataSampleSize):
         self.args.setLabelDataSampleSize(labelDataSampleSize)
+
+    @staticmethod
+    def writeArgumentsToJSON(fileName, args):
+        jvm.zingg.client.Arguments.writeArgumentsToJSON(fileName, args)
+
+class ClientOptions:
+
```
sonalgoyal, line 89: can we refer to the java constants here?

sonalgoyal (changes requested):

core/src/main/java/zingg/ZinggBase.java
navinrathore: done

python/phases/assessModel.py
navinrathore: removed

python/phases/assessModel.py
navinrathore: Done wherever possible.

python/phases/zingg.py
navinrathore: done

PR #319, opened 2022-06-08
new phase PeekModel

RavirajBaraiya:


client/src/main/java/zingg/client/ZinggOptions.java
```diff
@@ -14,6 +14,8 @@ public enum ZinggOptions {
     LINK("link"),
     GENERATE_DOCS("generateDocs"),
     UPDATE_LABEL("updateLabel"),
+    ASSESS_MODEL("assessModel"),
```
sonalgoyal, line 17: We don’t need assess model

core/src/main/java/zingg/PeekModel.java
```diff
@@ -0,0 +1,32 @@
...
+
+import zingg.client.ZinggClientException;
+import zingg.client.ZinggOptions;
+
+public class PeekModel extends ZinggBase{
+
+	protected static String name = "zingg.PeekModel";
+	public static final Log LOG = LogFactory.getLog(PeekModel.class); 
+
+    public PeekModel() {
```
sonalgoyal, line 14: Indentation is off

core/src/test/java/zingg/TestPeekModel.java
```diff
@@ -0,0 +1,33 @@
...
+		}
+    }
+
+    @Test
+    public void testOutput() throws Throwable{
+        PeekModel doc = new PeekModel();
+        doc.init(args, "");
+        doc.setArgs(args);
+        doc.execute();
+        doc.getMarkedRecords();
```
sonalgoyal, line 31: You need to validate the records - count etc

sonalgoyal (changes requested):

PR #320, opened 2022-06-08
zingg pip package artifacts

RavirajBaraiya:


package_zingg/setup.py
```diff
@@ -0,0 +1,26 @@
+from setuptools import *
+
+setup(
+    name='zingg',  
+    version='0.0.2',
```
sonalgoyal, line 5: Version will be 0.3.4

package_zingg/setup.py
```diff
@@ -0,0 +1,26 @@
...
+    include_package_data=True,
+    # packages=['zingg_jar'],
+    scripts=['zingg.py'] ,
+    author="Sonal Goyal",
+    author_email="sonalgoyal4@gmail.com",
+    description="Zingg.ai environment package",
```
sonalgoyal, line 11: Desc will be Python api for Zingg

package_zingg/zingg.py
```diff
@@ -0,0 +1,83 @@
...
+        return self.fd
+        
+class Pipe:
+    def __init__(self, name, format):
+        self.pipe = sc._jvm.zingg.client.pipe.Pipe()
+        self.pipe.setName(name)
+        self.pipe.setFormat(sc._jvm.zingg.client.pipe.Format.getPipeType(format))
+    def getPipe(self):
+        return self.pipe
+
```
sonalgoyal, line 81: How does the jar get packaged?

sonalgoyal (changes requested):

SubNader:
Hello @RavirajBaraiya @sonalgoyal my team is interested to try Zingg out. Will there be a pip package now that this is closed? Thanks!

sonalgoyal:
yes @SubNader - we are planning a pip package for the Zingg client. what kind of environment and data sizes do you have?

sonalgoyal:
we are targetting the week of 20th. if you want, you can also pick up the example at https://github.com/zinggAI/zingg/blob/main/api/python/FebrlExample.py and the api at https://github.com/zinggAI/zingg/blob/main/python/phases/zingg.py and work like standard pyspark code. you will need the latest zingg jar though, which we can provide if you have trouble compiling.

PR #326, opened 2022-06-13
API chnages issue part 2,4,5

RavirajBaraiya:


sonalgoyal (changes requested):
Keep all flavours of unit and execute, the wrapper method initandexec will be used in most cases, but in some cases we will not use it and use the atomic unit and execute

also did not understand the changes to febrl example

PR #332, opened 2022-06-14
formatting of help message

RavirajBaraiya:


sonalgoyal (changes requested):
Please build some kind of formatting so that we do not have to worry about this for nee phases for eg https://stackoverflow.com/questions/18672643/how-to-print-a-table-of-information-in-java

Issue #333, opened 2022-06-14
update febrl example with new py api changes


sonalgoyal:
update ClientOptions and Pipes.

PR #342, opened 2022-06-16
changed febrlExample according to new py api issue #333

RavirajBaraiya:


python/examples/FebrlExample.py
```diff
@@ -28,21 +29,21 @@
...
 #setting outputpipe in 'args'
-outputPipe = Pipe("result", "csv")
-outputPipe.addProperty("location","/tmp")
+outputPipe = CsvPipe("result")
+outputPipe.setLocation("/tmp")
 
 args.setOutput(outputPipe)
 
-options = ClientOptions(["--phase", "trainMatch",  "--conf", "dummy", "--license", "dummy", "--email", "xxx@yyy.com"])
+options = ClientOptions()
```
sonalgoyal, line 46: How will trainMatch be specified?

python/examples/FebrlExample.py
```diff
@@ -1,4 +1,5 @@
 from python.api.zingg import *
+from python.api.zinggPipes import *
```
sonalgoyal, line 2: Let us discuss this namespace

sonalgoyal (changes requested):
See comments

PR #346, opened 2022-06-17
Dockerfile changes for python packages and api support

navinrathore:
Python api support
Release 0.3.4

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,9 @@ USER root
 WORKDIR /zingg-0.3.3-SNAPSHOT
 RUN curl --location https://github.com/zinggAI/zingg/releases/download/v0.3.3/zingg-0.3.3-SNAPSHOT-spark-3.1.2.tar.gz | \
 tar --extract --gzip --strip=1 
-RUN chmod -R a+rwx /zingg-0.3.3-SNAPSHOT/models
-RUN chown -R 1001 /zingg-0.3.3-SNAPSHOT/models
+RUN pip install -r python/requirements.txt
+ENV PYTHONPATH /zingg-0.3.4-SNAPSHOT
+RUN chmod -R a+rwx /zingg-0.3.4-SNAPSHOT/models
+RUN chown -R 1001 /zingg-0.3.4-SNAPSHOT/models
 USER 1001
 


Dockerfile
```diff
@@ -8,7 +8,9 @@ USER root
 WORKDIR /zingg-0.3.3-SNAPSHOT
 RUN curl --location https://github.com/zinggAI/zingg/releases/download/v0.3.3/zingg-0.3.3-SNAPSHOT-spark-3.1.2.tar.gz | \
 tar --extract --gzip --strip=1 
-RUN chmod -R a+rwx /zingg-0.3.3-SNAPSHOT/models
-RUN chown -R 1001 /zingg-0.3.3-SNAPSHOT/models
+RUN pip install -r python/requirements.txt
```
sonalgoyal, line 11: will this install the zingg package as well?

Dockerfile
```diff
@@ -8,7 +8,9 @@ USER root
 WORKDIR /zingg-0.3.3-SNAPSHOT
```
sonalgoyal, line 8: wont these be 0.3.4?

sonalgoyal (changes requested):
0.3.4 and zingg pip package

Dockerfile
navinrathore: No.
Should it be? Hardly, any advantage.

navinrathore: no

navinrathore:
Yes. It should be 0.3.4 only.
Will not install zingg package.
Will revert this PR.
It is due to Auto merge during rebase issue.

Dockerfile
sonalgoyal: I think we should do a pip install of zingg....not sure if it adds much, but just for the sake of complete release being in docker.

Dockerfile
```diff
@@ -8,7 +8,9 @@ USER root
...
 RUN curl --location https://github.com/zinggAI/zingg/releases/download/v0.3.3/zingg-0.3.3-SNAPSHOT-spark-3.1.2.tar.gz | \
 tar --extract --gzip --strip=1 
-RUN chmod -R a+rwx /zingg-0.3.3-SNAPSHOT/models
-RUN chown -R 1001 /zingg-0.3.3-SNAPSHOT/models
+RUN pip install -r python/requirements.txt
+ENV PYTHONPATH /zingg-0.3.4-SNAPSHOT
```
sonalgoyal, line 12: why do we need pythonpath?

sonalgoyal (changes requested):
rm pypath
add pip install

Issue #336, opened 2022-06-15
New phase recommend 


sonalgoyal:
Documenter to only have model and data documentation. Field pages to only have field info- type etc.

move stopWords to models/recommendations/stopWords as csv files only

navinrathore:
* PR:  #348 


Ref. issue
*  #363 
      PR: #368

PR #348, opened 2022-06-20
Phase recommend

navinrathore:
new phase "recommend' added
Moved StopWordsXXX from Documenter to Recommender/Profiler
fixed relative location of col documents in model.flth


core/src/main/java/zingg/profiler/DataColProfiler.java
```diff
@@ -0,0 +1,42 @@
...
+	StopWordsProfiler stopWordsProfile;
+	JavaSparkContext ctx;
+	
+	public DataColProfiler(SparkSession spark, JavaSparkContext ctx, Arguments args) {
+		super(spark, args);
+		this.ctx = ctx;
+		stopWordsProfile = new StopWordsProfiler(spark, args);
+	}
+
+	public void process(Dataset<Row> data) throws ZinggClientException {
```
sonalgoyal, line 26: What is the use of process method?

core/src/main/java/zingg/profiler/DataProfiler.java
```diff
@@ -0,0 +1,48 @@
...
+		dataColProfiler = new DataColProfiler(spark, ctx, args);
+	}
+	
+	public void process() throws ZinggClientException {
+		try {
+			LOG.info("Data profiling starts");
+
+			try {
+				data = PipeUtil.read(spark, false, false, args.getData());
+				LOG.info("Read input data : " + data.count());
```
sonalgoyal, line 33: Please don’t print counts in info - as it is a performance overhead.

core/src/main/java/zingg/profiler/ProfilerBase.java
```diff
@@ -0,0 +1,25 @@
...
+class ProfilerBase {
+	protected SparkSession spark;
+	protected Arguments args;
+
+	public ProfilerBase(SparkSession spark, Arguments args) {
+		this.spark = spark;
+		this.args = args;
+	}
+
+	protected void checkAndCreateDir(String dirName) {
```
sonalgoyal, line 18: As we are using spark for writing, the folder creation would happen through it, no?

core/src/main/java/zingg/profiler/StopWordsProfiler.java
```diff
@@ -0,0 +1,52 @@
...
+public class StopWordsProfiler extends ProfilerBase {
+	protected static String name = "zingg.StopWordsProfiler";
+	public static final Log LOG = LogFactory.getLog(StopWordsProfiler.class);
+ 
+	public StopWordsProfiler(SparkSession spark, Arguments args) {
+		super(spark, args);
+		checkAndCreateDir(getStopWordsDir());
+	}
+
+	public void process(Dataset<Row> data) throws ZinggClientException {
```
sonalgoyal, line 27: Why this method?

core/src/main/java/zingg/profiler/StopWordsProfiler.java
```diff
@@ -0,0 +1,52 @@
...
+			data = data.select(split(data.col(fieldName), "\\s+").as("split"));
+			data = data.select(explode(data.col("split")).as("word"));
+			data = data.filter(data.col("word").notEqual(""));
+			data = data.groupBy("word").count().orderBy(desc("count"));
+			data = data.limit(Math.round(data.count()*args.getStopWordsCutoff()));
+		}
+		return data;
+	}
+
+	public String getStopWordsDir() {
```
sonalgoyal, line 49: Move to args

core/src/main/java/zingg/profiler/StopWordsProfiler.java
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	public void process(Dataset<Row> data) throws ZinggClientException {
+		//createStopWordsDocuments(data);
+	}
+
+	public void createStopWordsDocument(Dataset<Row> data, String fieldName, JavaSparkContext ctx) throws ZinggClientException {
+		String filenameCSV = getStopWordsDir() + fieldName;
+		data = findStopWords(data, fieldName);
+		PipeUtil.write(data, args, ctx, PipeUtil.getStopWordsPipe(args, filenameCSV));
```
sonalgoyal, line 34: Write with overwrite

core/src/main/java/zingg/profiler/DataProfiler.java
sonalgoyal: Why do we need the second try catch? Why not use the outer catch as pipe until prints already that it was not able to read a pipe

core/src/main/java/zingg/profiler/DataProfiler.java
```diff
@@ -0,0 +1,48 @@
...
+public class DataProfiler extends ProfilerBase {
+	protected static String name = "zingg.DataProfiler";
+
+	public static final Log LOG = LogFactory.getLog(DataProfiler.class);
+	private DataColProfiler dataColProfiler;
+	protected Dataset<Row> data;
+
+	public DataProfiler(SparkSession spark, JavaSparkContext ctx, Arguments args) {
+		super(spark, args);
+		data = spark.emptyDataFrame();
```
sonalgoyal, line 23: Why is an entity df needed?

sonalgoyal (changes requested):

sonalgoyal:
Add --column and generate only for that

core/src/main/java/zingg/profiler/DataProfiler.java
navinrathore: I think it is there from initial days when Model and Data documenters were combined. It was supposed to run until it is fatal.
However, removed now.

core/src/main/java/zingg/profiler/StopWordsProfiler.java
navinrathore: Done. For all times, in StopWordPipe.

core/src/main/java/zingg/profiler/StopWordsProfiler.java
navinrathore: done

core/src/main/java/zingg/profiler/StopWordsProfiler.java
navinrathore: Removed

core/src/main/java/zingg/profiler/ProfilerBase.java
navinrathore: Yes. It would create. removed the check and the function definition.

core/src/main/java/zingg/profiler/DataColProfiler.java
navinrathore: As there may be local col profiling in addition to that in other specific classes, if any. This function has been there when things were combined. In DataDocumenter, only this function is enough to call.

core/src/main/java/zingg/profiler/DataProfiler.java
navinrathore: Count() removed.
Outer catch removed. exception, if any, will be pass to higher in call stack.

navinrathore:
Removed an Reflection call.

sonalgoyal (commented):
rename all classes to recommender/package to recommend instead of profiler

core/src/main/java/zingg/profiler/ProfilerBase.java
sonalgoyal: i still see this?

core/src/main/java/zingg/profiler/DataColProfiler.java
```diff
@@ -0,0 +1,46 @@
...
+				if(Arrays.asList(data.schema().fieldNames()).contains(args.getColumn())) {
+					stopWordsProfile.createStopWordsDocument(data, args.getColumn(), ctx);
+				} else {
+					LOG.info("An invalid column name - " + args.getColumn() + " entered. Please provide valid column name.");
+				}
+			} else {
+				LOG.info("Please provide '--column <columnName>' option at command line to generate stop words for that column.");
+			}
+		} else {
+			LOG.info("No Stop Words document generated");
```
sonalgoyal, line 43: it is a recommendation - please update error message

core/src/main/java/zingg/profiler/DataProfiler.java
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+
+import zingg.client.Arguments;
+import zingg.client.ZinggClientException;
+import zingg.util.PipeUtil;
+
+public class DataProfiler extends ProfilerBase {
+	protected static String name = "zingg.DataProfiler";
```
sonalgoyal, line 15: are we using name anywhere?

core/src/main/java/zingg/profiler/StopWordsProfiler.java
```diff
@@ -0,0 +1,43 @@
...
+	public void createStopWordsDocument(Dataset<Row> data, String fieldName, JavaSparkContext ctx) throws ZinggClientException {
+		String filenameCSV = args.getStopWordsDir() + fieldName;
+		data = findStopWords(data, fieldName);
+		PipeUtil.write(data, args, ctx, PipeUtil.getStopWordsPipe(args, filenameCSV));
+	}
+
+	private Dataset<Row> findStopWords(Dataset<Row> data, String fieldName) {
+		LOG.debug("Field: " + fieldName);
+		if(!data.isEmpty()) {
+			data = data.select(split(data.col(fieldName), "\\s+").as("split"));
```
sonalgoyal, line 35: column names created by Zingg are defined centrally in ColNames - all of them have z_ prefix

core/src/main/java/zingg/profiler/StopWordsProfiler.java
```diff
@@ -0,0 +1,43 @@
...
+			data = data.select(explode(data.col("split")).as("word"));
+			data = data.filter(data.col("word").notEqual(""));
+			data = data.groupBy("word").count().orderBy(desc("count"));
```
sonalgoyal, line 38: can we filter the data based on the cutoff and then orderby? it may be faster

core/src/main/resources/model.ftlh
```diff
@@ -18,7 +18,7 @@
   <tr>
     <th class="border-right border-white" style="width: 160px;" >Cluster</th>
     <#list 3 ..< numColumns as entityIndex>
-    <th class="border-right border-white" > <a href="docs/${columns[entityIndex]}.html"> ${columns[entityIndex]!} </a></th>
+    <th class="border-right border-white" > <a href="${columns[entityIndex]}.html"> ${columns[entityIndex]!} </a></th>
```
sonalgoyal, line 21: do we still need the links? what are we showing here when we click?

core/src/test/java/zingg/profiler/TestStopWordsProfiler.java
```diff
@@ -0,0 +1,90 @@
+package zingg.profiler;
```
sonalgoyal, line 1: the main test here should not be reading/writing files but if we are generating the write stop words. just build a dataset in memory and use that as your dataset for testing?

client/src/main/java/zingg/client/Client.java
```diff
@@ -92,6 +92,10 @@ else if (args.getJobId() != -1) {
 			String j = options.get(ClientOptions.SHOW_CONCISE).value;
 			args.setShowConcise(Boolean.valueOf(j));
 		}
+		if (options.get(ClientOptions.COLUMN)!= null) {
```
sonalgoyal, line 95: can we have a test here to see this is getting set correctly?

sonalgoyal (changes requested):
added comments, please check

client/src/main/java/zingg/client/Arguments.java
```diff
@@ -108,6 +108,7 @@ public class Arguments implements Serializable {
 	boolean showConcise = false;
 	float stopWordsCutoff = 0.1f;
 	long blockSize = 100L;
+	String column = "";
```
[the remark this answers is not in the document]
navinrathore: Other option is "null".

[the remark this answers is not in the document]
navinrathore: Set it uninitialized that is nothing but a null.

core/src/main/java/zingg/profiler/ProfilerBase.java
```diff
@@ -0,0 +1,25 @@
+package zingg.profiler;
```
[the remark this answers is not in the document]
navinrathore: Removed the class

core/src/main/java/zingg/util/PipeUtil.java
```diff
@@ -336,6 +336,7 @@ public static Pipe getStopWordsPipe(Arguments args, String fileName) {
 		p.setFormat(Format.CSV);
 		p.setProp(FilePipe.HEADER, "true");
```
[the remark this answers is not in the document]
navinrathore: Added testcase

core/src/main/java/zingg/profiler/StopWordsProfiler.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+
+import zingg.client.Arguments;
+import zingg.client.ZinggClientException;
+import zingg.util.PipeUtil;
+
+public class StopWordsProfiler extends ProfilerBase {
+	protected static String name = "zingg.StopWordsProfiler";
```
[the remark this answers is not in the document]
navinrathore: Removed.

core/src/main/java/zingg/profiler/DataColProfiler.java
navinrathore: Updated to "No stopwords generated"

core/src/main/java/zingg/profiler/DataProfiler.java
navinrathore: DataProfiler and DataColProfiler are the two classes. Removed DataColProfiler and moved applicable stuff in the class DataProfiler,

core/src/main/java/zingg/profiler/StopWordsProfiler.java
navinrathore: Made them local consts only. They are not reusable column names.

core/src/main/java/zingg/profiler/StopWordsProfiler.java
navinrathore: Stopwords chosen whose count is greaterThan(sum(col(count))* cuttoff); 
No more orderBy()
**Still Need review of function findStopWords() as the result size may be UN-predictable. Largely depend on how many words are there in dataset"**

client/src/main/java/zingg/client/Client.java
navinrathore: Yes. Added the testcase.

core/src/test/java/zingg/profiler/TestStopWordsProfiler.java
navinrathore: Removed such tests. added one wherein dataset is created programmatically.

core/src/main/java/zingg/profiler/ProfilerBase.java
navinrathore: ProfilerBase.java itself has been removed.

navinrathore:
#336

navinrathore:
[Everywhere names comprising profiler changed to that of recommender(individual commit)](https://github.com/zinggAI/zingg/pull/348/commits/18575f060f42c37ef459a865f4574848fe925908)

core/src/test/java/zingg/recommender/TestStopWordsRecommender.java
```diff
@@ -0,0 +1,90 @@
...
+
+		Dataset<Row> dataset = createDFWithGivenStopWords();
+
+		args.setStopWordsCutoff(0.1f);
+		Dataset<Row> stopWords = recommender.findStopWords(dataset, COL_STOPWORDS);
+		stopWords.show();
+	}
+	/* creates a dataframe for given words and their frequency*/
+	public Dataset<Row> createDFWithGivenStopWords() {
+		Map<String, Integer> map = Stream.of(new Object[][] {
```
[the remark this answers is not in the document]
sonalgoyal: The data creation here is leading to a lot of extra loops and object creation. Use of a hash when 

I feel a simpler logic would be
- define a class word, count (the, 44)
- define a list of words. just add all the words one by one. 
- define wordDistribution - int[][] row is NO_OF_RECORDS, each column represents a word
- fill wordDistribution per word in the list of words
- iterate over the wordDist array and join the strings by looking up the col index. 

Also define structType once and reuse?

sonalgoyal (changes requested):

Issue #349, opened 2022-06-20
Null Pointer exception occures when "DoubleType" column has string data


navinrathore:
**Describe the bug**
NullPointerException is thrown when a column of DoubleType has non-double data say string. e.g. if price includes string '$' or 'GBP'.

**To Reproduce**
The price column of GoogleProduct.csv has some records that include string "GBP".
Run scripts/zingg.sh --phase train --conf examples/amazon-google/config.json

![Screenshot_20220617_113525](https://user-images.githubusercontent.com/19220734/174586996-e2809c63-5882-41b5-a416-b949f716c6ab.png)

PR #350, opened 2022-06-21
Null Pointer check in "Range" hash functions

navinrathore:
* Handled 'Null' input value in Range set of hash functions. Added testcases also for same.
* All other functions have the null chech. Test cases were also present.
* List of new hash functions recently added
**LessThanZeroInt,LessThanZeroDbl,RangeDbl,RangeInt,,TruncateDouble,TrimLastDigitsInt,TrimLastDigitsDbl**

### DIFF
--- a/core/src/main/java/zingg/hash/RangeDbl.java
+++ b/core/src/main/java/zingg/hash/RangeDbl.java
@@ -17,7 +17,7 @@ public class RangeDbl extends HashFunction implements UDF1<Double, Integer> {
 	@Override
 	public Integer call(Double field) {
 		int withinRange = 0;
-		if (field >= lowerLimit && field < upperLimit) {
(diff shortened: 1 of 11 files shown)


navinrathore:
Ref:  Null Pointer exception occures when "DoubleType" column has string data #349

sonalgoyal (changes requested):
the apply method - will it work or will it fail? I dont see the changes there..?

navinrathore:
> the apply method - will it work or will it fail? I dont see the changes there..?
For the issue, mentioned in this bug is solved by this fix. the fix is in consonance with other hash functions implementation.

The casting issue/message may have come in some other scenario/setup. The getAs() function is overloaded. It is supposed to return Double only. Aberrant scenario we have to find. We may raise a bug for that too!

Issue #308, opened 2022-06-02
create zingg pip python package 


RavirajBaraiya:
create Zingg pip python package which includes java jar and scripts.

When you call the final Zingg package using: pip install zingg, it should be able to invoke Java classes.

RavirajBaraiya:
[Zingg pip python package.docx](https://github.com/zinggAI/zingg/files/8824748/Zingg.pip.python.package.docx)

sonalgoyal:
Leaving this open for testing and building a script.

RavirajBaraiya:
[Zingg pip python package.docx](https://github.com/zinggAI/zingg/files/8824763/Zingg.pip.python.package.docx)

sonalgoyal:
@RavirajBaraiya - please can you checkin all the artifacts created - setup script, init py, manifest etc

sonalgoyal:
please add them to the relevnt locations under python folder so that we can run the commands and test this out.

RavirajBaraiya:
after pr #356 
after installing zingg, when I am running any python file or phase using zingg.sh, it is giving Warn:

/home/raviraj/.local/bin/zingg.sh: line 20: cd: /home/raviraj/.local/bin/../config: No such file or directory
sed: can't read /home/raviraj/zingg/zingg.conf: No such file or directory
/home/raviraj/.local/bin/zingg.sh: line 35: ENVIRONMENT_VARS[${key}]: bad array subscript

sonalgoyal:
i am getting this error after running sdist and then installing
white_tick
eyes
raised_hands





[23:48](https://zinggai.slack.com/archives/C03DLRXTL05/p1656008310011139)
pip install dist/*.tar.gz
Traceback (most recent call last):
  File "/usr/local/bin/pip", line 6, in <module>
    from pip._internal.cli.main import main
ImportError: No module named main

sonalgoyal:
couple of issues with the package
- original jar is getting added
- deps is a separate folder at the same level as zingg

PR #356, opened 2022-06-21
zingg pip package artifacts modification

RavirajBaraiya:


RavirajBaraiya:
create Zingg pip python package issue #308

PR #365, opened 2022-06-24
new zinggExamples package issue#361

RavirajBaraiya:


sonalgoyal (commented):
Can we keep common files needed for api and example packages under oython and use them as necessary so we don’t make extra files?

sonalgoyal:
closing this pr. let us have one package with 2 examples.

Issue #363, opened 2022-06-24
test data documenter breaking


sonalgoyal:
break in zingg.documenter.TestDataDocumenter.testIfDataDocumenterGeneratedDocFile(TestDataDocumenter.java:48)
![Screenshot from 2022-06-23 23-36-43](https://user-images.githubusercontent.com/309387/175468903-8df9161d-af7f-4515-98b4-54426a46046b.png)

navinrathore:
DataDocumenter.java and TestDataDocumenter.java have been removed. They do not have any significant functionality.

PR #367, opened 2022-06-27
DataDocumenter and TestDataDocumenter removed

navinrathore:
DataDocumenter removed as it does not have fall into category of model documentation. Currently, generated data.html has only details of field definitions.


navinrathore:
#363 test data documenter breaking

Issue #351, opened 2022-06-21
Exception while running zingg with amazon-google dataset


navinrathore:
**Describe the bug**
Below exception occurs when match/findTrainignData is run for amazon-google dataset

```
 2022-06-21 08:01:09,486 [main] WARN  zingg.client.Client - Apologies for this message. Zingg has encountered an error. 
Job aborted due to stage failure: Task 1 in stage 145.0 failed 1 times, most recent failure: Lost task 1.0 in stage 145.0 (TID 233) (192.168.29.65 executor driver): 
org.apache.spark.SparkException: 
Failed to execute user defined function(UDFRegistration$$Lambda$3374/1427732162: (double, double) => double)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anon$1.hasNext(WholeStageCodegenExec.scala:755)
	at org.apache.spark.sql.execution.columnar.DefaultCachedBatchSerializer$$anon$1.next(InMemoryRelation.scala:87)
	at org.apache.spark.sql.execution.columnar.DefaultCachedBatchSerializer$$anon$1.next(InMemoryRelation.scala:79)
	at scala.collection.Iterator$$anon$10.next(Iterator.scala:459)
	at org.apache.spark.storage.memory.MemoryStore.putIterator(MemoryStore.scala:222)
	at org.apache.spark.storage.memory.MemoryStore.putIteratorAsValues(MemoryStore.scala:299)
	at org.apache.spark.storage.BlockManager.$anonfun$doPutIterator$1(BlockManager.scala:1423)
	at org.apache.spark.storage.BlockManager.org$apache$spark$storage$BlockManager$$doPut(BlockManager.scala:1350)
	at org.apache.spark.storage.BlockManager.doPutIterator(BlockManager.scala:1414)
	at org.apache.spark.storage.BlockManager.getOrElseUpdate(BlockManager.scala:1237)
	at org.apache.spark.rdd.RDD.getOrCompute(RDD.scala:384)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:335)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
[...]

PR #369, opened 2022-06-28
Double similarity function - null pointer exception

navinrathore:
* fixed config.json as MatchType NUMERIC_WITH_UNITS changed to NUMBER_WITH_UNITS
* 'null' checks in DoubleSimilarityFunction. Also, checked for Double.NAN

### DIFF
--- a/core/src/test/java/zingg/similarity/function/TestDoubleSimilarityFunction.java
+++ b/core/src/test/java/zingg/similarity/function/TestDoubleSimilarityFunction.java
@@ -1,0 +1,51 @@
+package zingg.similarity.function;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+public class TestDoubleSimilarityFunction {
+	
+	
+	@Test
+	public void testFirstNumIsNull() {
+		DoubleSimilarityFunction isNull = new DoubleSimilarityFunction();
+		assertEquals(1d, isNull.call(null, 1d));
+	}
+
+	@Test
+	public void testFirstNumIsNAN() {
+		DoubleSimilarityFunction isNull = new DoubleSimilarityFunction();
+		assertEquals(1d, isNull.call(Double.NaN, 1d));
+	}
+
+	@Test
+	public void testSecondNumIsNull() {
+		DoubleSimilarityFunction isNull = new DoubleSimilarityFunction();
+		assertEquals(1d, isNull.call(1d, null));
+	}
+
+	@Test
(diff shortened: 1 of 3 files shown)


navinrathore:
Tested with febrl standard dataset and another updated with Double field/values
with amazon-google dataset
with iTunes-amazon dataset

navinrathore:
Noticed that, for amazon-google dataset, header for google.csv file is not compatible with names in schema. Though, zingg seems working fine but below is the message.
```
2022-06-28 11:39:15,950 [main] WARN  zingg.util.PipeUtil - Reading Pipe [name=google, format=CSV, preprocessors=null, props={location=examples/amazon-google/GoogleProducts.csv, delimiter=,, header=true}, schema=StructType(StructField(id,StringType,false), StructField(title,StringType,true), StructField(description,StringType,true), StructField(manufacturer,StringType,true), StructField(price,DoubleType,true))]
 2022-06-28 11:39:16,248 [Executor task launch worker for task 1.0 in stage 10.0 (TID 39)] WARN  org.apache.spark.sql.catalyst.csv.CSVHeaderChecker - CSV header does not conform to the schema.
 Header: id, name, description, manufacturer, price
 Schema: id, title, description, manufacturer, price
Expected: title but found: name
CSV file: file:///home/navin/workDir/zingg-1/examples/amazon-google/GoogleProducts.csv

```

navinrathore:
* #351

core/src/test/java/zingg/similarity/function/TestDoubleSimilarityFunction.java
```diff
@@ -0,0 +1,51 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+public class TestDoubleSimilarityFunction {
+	
+	
+	@Test
+	public void testFirstNumIsNull() {
+		DoubleSimilarityFunction isNull = new DoubleSimilarityFunction();
```
sonalgoyal, line 12: name of variable?

sonalgoyal (changes requested):
- Add tests for some doubles.
- Rename variable

core/src/test/java/zingg/similarity/function/TestDoubleSimilarityFunction.java
navinrathore: changed to simFn

navinrathore: Added some tests

PR #370, opened 2022-06-29
changed 'Format' type from Enum to Class to support new formats by default

navinrathore:
* changes made for **Format** class to behave as close as possible earlier enum avatar.
* **PipeFactory** class seems redundant as of now. Particularly, its register() function. Therefore, its getPipe() functions has hardcoded format types. switch statement supports enum or constants only.

### DIFF
--- a/client/src/main/java/zingg/client/pipe/CassandraPipe.java
+++ b/client/src/main/java/zingg/client/pipe/CassandraPipe.java
@@ -10,4 +10,7 @@ public class CassandraPipe extends Pipe{
 	public static final String CLUSTER_BY="clusterBy";
 	public static final String INDEX_BY="indexBy";
 
+	public CassandraPipe() {
+		setFormat(Format.CASSANDRA);
+	}
 }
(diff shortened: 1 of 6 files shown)


navinrathore:
* Tested with mysql apart from csv. Parquet is internally used at multiple place.
* Tried ORC format as well. it could read file.
* In fact, tested with CSV, Parquet removed from the Format class and things worked properly. This suggests it should work any supported format.
* For bad formats. the following error message prints from thrown exception in addition to local message.
```
2022-06-29 11:46:29,053 [main] WARN  zingg.util.PipeUtil - 
Failed to find data source: csv2. Please find packages at http://spark.apache.org/third-party-projects.html
 2022-06-29 11:46:30,786 [main] WARN  zingg.client.Client - 
Apologies for this message. Zingg has encountered an error. Could not read data.

```
the actual thrown exception is below. (not printed).
```
java.lang.ClassNotFoundException: Failed to find data source: badFormat. Please find packages at http://spark.apache.org/third-party-projects.html
	at org.apache.spark.sql.execution.datasources.DataSource$.lookupDataSource(DataSource.scala:692)
```

navinrathore:
After the enum to class change, it is now mandatory for Pipe to have not-null format. Shall be put this restriction somehow?
InMemoryPipe, JDBCPipe have required such change.

sonalgoyal (changes requested):
Do you think we could remove Format class altogether and just keep it as Pipe and String format?

navinrathore:
> Do you think we could remove Format class altogether and just keep it as Pipe and String format?

Yes. we can. Though, It has got references/usage at multiple places.

Issue #347, opened 2022-06-17
python api changes


sonalgoyal:
add phase specific methods
inmemorypipe to accept pandas df

sonalgoyal:
constants in pipes to move to java and referenced here

sonalgoyal:
def getDfFromDs(self, data):
        return DataFrame(data, sqlContext) 

this gives an error /databricks/spark/python/pyspark/sql/dataframe.py:129: UserWarning: DataFrame constructor is internal. Do not directly use it.
  warnings.warn("DataFrame constructor is internal. Do not directly use it.")

sonalgoyal:
getUnMarkedRecords - rename to getUnmarkedRecords

navinrathore:
> getUnMarkedRecords - rename to getUnmarkedRecords

* PR #372

navinrathore:
> getUnMarkedRecords - rename to getUnmarkedRecords

* PR #372

navinrathore:
> def getDfFromDs(self, data): return DataFrame(data, sqlContext)
> 
> this gives an error /databricks/spark/python/pyspark/sql/dataframe.py:129: UserWarning: DataFrame constructor is internal. Do not directly use it. warnings.warn("DataFrame constructor is internal. Do not directly use it.")

* PR #373

PR #373, opened 2022-07-01
To fix Databrics UserWarning: DataFrame constructor is internal...

navinrathore:
Used SparkSession object instead. SqlContext is deprecated. [Code link](https://spark.apache.org/docs/latest/api/python/_modules/pyspark/sql/dataframe.html)

### DIFF
--- a/python/zingg/zingg.py
+++ b/python/zingg/zingg.py
@@ -51,9 +51,10 @@ class Zingg:
     def getUnsureMarkedRecordsStat(self):
         return self.client.getUnsureMarkedRecordsStat(self.getMarkedRecords())
     def getDfFromDs(self, data):
-        return DataFrame(data, sqlContext)
+        return DataFrame(data, spark)
     def getPandasDfFromDs(self, data):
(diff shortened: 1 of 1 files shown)


navinrathore:
Regarding replacing toPandas(), this funciton gives following error
```
  File "/home/navin/workDir/zingg-1/python/zingg/zingg.py", line 58, in getPandasDfFromDs
    return self.getDfFromDs(data).toPandas()
  File "/home/navin/softlibs/spark-3.1.2-bin-hadoop2.7/python/lib/pyspark.zip/pyspark/sql/pandas/conversion.py", line 67, in toPandas
AttributeError: 'SparkSession' object has no attribute '_conf'
```
As pyspark is built from JVM bsed Dataframe. [internally many JVM apis are called on it], seems there is limitations such as missing toPandas().
Therefore, used pd.DataFrame() to build pandas dataframe.

* Needless to say pandas dataframe are meant only for small sets of records

navinrathore:
* fixes a part of https://github.com/zinggAI/zingg/issues/347

PR #374, opened 2022-07-01
Added specific pipes property constants

navinrathore:


navinrathore:
* takes care of a part of https://github.com/zinggAI/zingg/issues/347

python/zingg/pipes/pipes.py
```diff
@@ -6,46 +6,48 @@
...
+        Pipe.addProperty(self, FilePipe.DELIMITER, delimiter)
     def setLocation(self, location):
         Pipe.addProperty(self, FilePipe.LOCATION, location)
 
 class BigQueryPipe(Pipe):
     def __init__(self,name):
-        Pipe.__init__(self, name, "bigquery")
-        Pipe.addProperty(self, "viewsEnabled", "true")
+        Pipe.__init__(self, name, Format.BIGQUERY.type())
+        Pipe.addProperty(self, JBigQueryPipe.VIEWS_ENABLED, "true")
```
sonalgoyal, line 24: can we add this property by default?

python/zingg/pipes/pipes.py
```diff
@@ -6,46 +6,48 @@
 
 Format = jvm.zingg.client.pipe.Format
 FilePipe = jvm.zingg.client.pipe.FilePipe
+JSnowflakePipe = jvm.zingg.client.pipe.SnowflakePipe
+JBigQueryPipe = jvm.zingg.client.pipe.BigQueryPipe
 
 class CsvPipe(Pipe):
     def __init__(self, name):
         Pipe.__init__(self, name, Format.CSV.type())
         Pipe.addProperty(self, FilePipe.HEADER,"true")
```
sonalgoyal, line 15: this is wrong - we can not set header default true. we have to let the user set the values.

sonalgoyal (changes requested):

python/zingg/pipes/pipes.py
navinrathore: I've removed it now.

python/zingg/pipes/pipes.py
navinrathore: In Java app, there are no settings for BigQueryPipe,

navinrathore:
> please add newlines between python methods.
Already in the code base
https://github.com/zinggAI/zingg/blob/main/python/zingg/pipes/pipes.py
through the below PR (for same comment here) 
https://github.com/zinggAI/zingg/pull/376#discussion_r913627667

python/zingg/pipes/pipes.py
navinrathore: Property is no more set here.

Issue #252, opened 2022-05-11
Add new config files for "Amazon-google" dataset involving stop word removal.


navinrathore:
**Description**
* For "Amazon-google" dataset, add new config with stopwords attributes
* Update documentation for, giving this example, how one can work with stop words.

sonalgoyal:
It is not a new model, just add a new config and add documentation of first using config without stop words settings to generate, then select few files and edit them and put them in the main folder and create an updated config.

PR #392, opened 2022-07-13
added stopword functionality in zingg FieldDefinition

RavirajBaraiya:
added setStopWordsCutoff method in zingg.Arguments class
added setStopWords method in zingg.FeildDefinition class
added stopword parameter in one field of amazon-google and iTunes-amazon examples


RavirajBaraiya:
issue #340

RavirajBaraiya:
issue #252 
added new config files for "Amazon-google" dataset involving stop word removal.

examples/stopWords.csv
```diff
@@ -0,0 +1,16 @@
+StopWord
```
sonalgoyal, line 1: move this file into amazon google and update code and config accordingly

sonalgoyal (changes requested):
move the stop words file under the relevant example folder and use from there.

examples/stopWords.csv
RavirajBaraiya: moved csv to amazon-google

Issue #338, opened 2022-06-15
tests for python


sonalgoyal:
add febrl example test

vikasgupta78:
Added readme on how to run the python test suite.

above was run in a way that it was not picking the test file

./scripts/zingg.sh --run   test/pythonTest.py

Pull request #588 commit 8d1c8d6a

PR #395, opened 2022-07-14
added python script to run all with febrl example python unittest

RavirajBaraiya:
issue #338 

to run all python unittests
cmd: zingg.sh run python_test/pythonTest.py


sonalgoyal (changes requested):
rename python_test to test
can we reuse the csv under examples?

RavirajBaraiya:
renamed python_test to test
renamed test_febrl.py to testFebrl.py
reused the csv under examples

PR #399, opened 2022-07-15
recommender phase issue #336

RavirajBaraiya:


core/src/main/java/zingg/recommender/StopWordsRecommender.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+
+import zingg.client.Arguments;
+import zingg.client.ZinggClientException;
+import zingg.util.PipeUtil;
+
+public class StopWordsRecommender {
+	private static final String COL_COUNT = "count";
```
sonalgoyal, line 21: move these to ColNames class

sonalgoyal (changes requested):
new columns for stop words should be moved to ColNames file and have z_ appended.

core/src/main/java/zingg/recommender/StopWordsRecommender.java
RavirajBaraiya: added columns for stop words to ColNames file with z_ appended.

PR #438, opened 2022-08-30
adding session template

Akash-R-7:
For documenter package

### DIFF
--- a/core/src/main/java/zingg/documenter/DataDocumenter.java
+++ b/core/src/main/java/zingg/documenter/DataDocumenter.java
@@ -15,6 +15,7 @@ import org.apache.spark.sql.types.StructField;
 import zingg.client.Arguments;
 import zingg.client.ZFrame;
 import zingg.client.ZinggClientException;
+import zingg.util.PipeUtilBase;
 
 public class DataDocumenter<S,D,R,C,T> extends DocumenterBase<S,D,R,C,T> {
 	protected static String name = "zingg.DataDocumenter";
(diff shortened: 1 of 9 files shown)


core/src/main/java/zingg/documenter/DataDocumenter.java
```diff
@@ -24,17 +25,18 @@
...
 	}
 	
+
 	public void process() throws ZinggClientException {
 		try {
 			LOG.info("Data document generation starts");
 
 			try {
-				data = PipeUtil.read(spark, false, false, args.getData());
+				data = PipeUtilBase.read(session, 0, false, args.getData());
```
sonalgoyal, line 39: this has to be getPipeUtil() - see trainer etc

sonalgoyal: pipeutils already have session info - i don't think we need to pass them. again check existing code

>

Akash-R-7: Yeah sorry I missed that function name there.

core/src/main/java/zingg/util/PipeUtilBase.java
```diff
@@ -20,6 +20,9 @@ public  ZFrame<D, R, C> readInternal(boolean addLineNo,
 	public  ZFrame<D, R, C> read(boolean addLineNo, int numPartitions,
 			boolean addSource, Pipe... pipes);
 
+	public  ZFrame<D, R, C> read(S session, int numPartitions,
```
sonalgoyal: this is not needed...we are creating concrete classes which have the session

sonalgoyal (changes requested):
minor changes - please do. else all looks good

PR #478, opened 2022-12-30
Generics

vikasgupta78:
spark hash function changes


core/src/main/java/zingg/spark/hash/SparkIdentityString.java
```diff
@@ -20,18 +20,34 @@ public SparkIdentityString() {
...
+    }
+
+    @Override
+    public Object apply(Row ds, String column) {
+        return call((String) getAs(ds, column));
+   }
+
+
+    @Override
+    public Object getAs(Dataset<Row> df, Row r, String column) {
```
sonalgoyal, line 43: arent we doing unsupportedexcptn?

sonalgoyal (changes requested):
pls check comment

PR #481, opened 2023-01-02
update json hash function list

vikasgupta78:


assembly/pom.xml
```diff
@@ -11,12 +11,22 @@
 <dependencies>
 		<dependency>
 			<groupId>zingg</groupId>
-			<artifactId>core</artifactId>
+			<artifactId>common_zingg_core</artifactId>
```
sonalgoyal, line 14: have some thoughts on naming.

PR #482, opened 2023-01-09
Junit hash functions - should be tested on base classes

vikasgupta78:


common/core/src/test/java/zingg/hash/TestLessThanZeroDbl.java
```diff
@@ -9,30 +9,30 @@ public class TestLessThanZeroDbl {
```
sonalgoyal, line 9: are we missing the not a number test here? I see it in the spark one which was deleted(spark_zingg/core/src/test/java/zingg/spark/hash/TestSparkLessThanZeroDbl.java) ?

sonalgoyal (changes requested):

common/core/src/test/java/zingg/hash/TestLessThanZeroDbl.java
vikasgupta78: added in latest commit, thanks

PR #491, opened 2023-02-27
Stop words recommender working

vikasgupta78:
Changes to make Stop words recommender working


spark_zingg/core/src/main/java/zingg/spark/SparkStopWordsRecommender.java
```diff
@@ -0,0 +1,60 @@
...
+	public static final Log LOG = LogFactory.getLog(SparkStopWordsRecommender.class);
+
+	public SparkStopWordsRecommender(Context<SparkSession, Dataset<Row>, Row, Column,DataType> context,Arguments args) {
+		super(context,args);
+	}
+	
+	@Override
+	public ZFrame<Dataset<Row>, Row, Column> findStopWords(ZFrame<Dataset<Row>, Row, Column> zDF, String fieldName) {
+		LOG.debug("Field: " + fieldName);
+
```
sonalgoyal, line 42: the logic of this method should be in the common class

common/core/src/main/java/zingg/recommender/StopWordsRecommender.java
```diff
@@ -55,23 +56,8 @@ public void createStopWordsDocuments(ZFrame<D,R,C> data, String fieldName) throw
 		} else {
 			LOG.info("No stopwords generated");
 		}
-		*/
+		
  	}
 
-	public ZFrame<D,R,C> findStopWords(ZFrame<D,R,C> data, String fieldName) {
```
sonalgoyal, line 61: see other comment. this base class should have all the logic for building stop words so that we can use across snow and spark.

sonalgoyal (changes requested):
please see changes

spark_zingg/core/src/main/java/zingg/spark/SparkStopWordsRecommender.java
vikasgupta78: i have tried to fix it in commit 771ae6a, please have a look. let me know if more improvement is needed

common/core/src/main/java/zingg/recommender/StopWordsRecommender.java
vikasgupta78: i have tried to fix it in commit 771ae6a, please have a look. let me know if more improvement is needed

vikasgupta78:
i have tried to fix it in commit 771ae6a, please have a look. let me know if more improvement is needed

PR #499, opened 2023-03-02
Stop words preprocessor now works

vikasgupta78:


vikasgupta78:
creating new with junits

Issue #558, opened 2023-03-31
link phase not working


vikasgupta78:
link phase giving error

./scripts/zingg.sh --phase  link --conf examples/amazon-google/config.json

 2023-03-31 17:30:55,864 [main] INFO  zingg.spark.core.model.SparkModel - threshold while predicting is 0.5
 2023-03-31 17:30:56,240 [main] DEBUG zingg.spark.core.similarity.SparkTransformer - transforming dataset for VectorValueExtractor
 2023-03-31 17:30:56,370 [main] DEBUG zingg.common.core.executor.Matcher - Found dupes 0
 org.apache.spark.sql.AnalysisException: Reference 'z_cluster' is ambiguous, could be: z_cluster, z_cluster.
	at org.apache.spark.sql.catalyst.expressions.package$AttributeSeq.resolve(package.scala:363)
	at org.apache.spark.sql.catalyst.plans.logical.LogicalPlan.resolve(LogicalPlan.scala:122)
	at org.apache.spark.sql.catalyst.plans.logical.LogicalPlan.resolveQuoted(LogicalPlan.scala:133)
	at org.apache.spark.sql.Dataset.resolve(Dataset.scala:262)
	at org.apache.spark.sql.Dataset.col(Dataset.scala:1359)
	at zingg.spark.client.SparkFrame.col(SparkFrame.java:110)
	at zingg.spark.client.SparkFrame.col(SparkFrame.java:18)
	at zingg.common.core.util.DSUtil.alignLinked(DSUtil.java:123)
	at zingg.common.core.executor.Linker.writeOutput(Linker.java:44)
	at zingg.common.core.executor.Matcher.execute(Matcher.java:140)
	at zingg.common.client.Client.execute(Client.java:243)
	at zingg.common.client.Client.mainMethod(Client.java:182)
	at zingg.spark.client.SparkClient.main(SparkClient.java:65)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.spark.deploy.JavaMainApplication.start(SparkApplication.scala:52)
[...]

PR #562, opened 2023-04-04
issue #558 linker not working

vikasgupta78:
linker issue fixed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,8 @@
-FROM docker.io/bitnami/spark:3.1.2
+#FROM docker.io/bitnami/spark:3.1.2
+FROM apache/spark-py:v3.1.3
+USER 0
+RUN apt-get update && \
+	apt install -y curl vim 
 ENV SPARK_MASTER local[*]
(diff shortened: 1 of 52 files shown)


spark/client/src/main/java/zingg/spark/client/SparkFrame.java
```diff
@@ -130,7 +130,12 @@ public ZFrame<Dataset<Row>, Row, Column> dropDuplicates(String c, String... d) {
     public ZFrame<Dataset<Row>, Row, Column> drop(String c) {
```
sonalgoyal, line 130: invoke drop(Column) here instead of going to spark api directly. then code is in one place only

vikasgupta78: i don't think that would helps as:
1. To invoke drop(Column) we would still use functions.col or df.col to get column object from column name , so dependency on spark is not reduced.
2. Currently spark frame works as a adapter over data set where we delegate most of the methods to the data set object. This is exactly what we are doing with both drop methods. Now introducing logic in one of the drop methods instead of simple delegation does not make much sense

Issue #560, opened 2023-04-03
Install number


vikasgupta78:
Install number should be generated if not already there when client runs.

This should go as part of telemetry

If permission to create not there than just log but don't fail

vikasgupta78:
fixed in pull request #566 commit 320666f6

PR #566, opened 2023-04-12
client id to be generated issue #560

vikasgupta78:
client id to be generated issue for analytics #560


sonalgoyal:
Will implement this  potentially differently. Closing this

PR #619, opened 2023-06-29
Session wrapper and license interface

vikasgupta78:
Session wrapper and license interface

### DIFF
--- a/common/client/src/main/java/zingg/common/client/license/ILicenseValidator.java
+++ b/common/client/src/main/java/zingg/common/client/license/ILicenseValidator.java
@@ -1,0 +1,8 @@
+package zingg.common.client.license;
+
+public interface ILicenseValidator {
+
+	
+	public boolean validate();
+	
+}
--- a/common/client/src/main/java/zingg/common/client/license/ILicenseValidator.java
+++ b/common/client/src/main/java/zingg/common/client/license/ILicenseValidator.java
@@ -2,6 +2,7 @@ package zingg.common.client.license;
 
 public interface ILicenseValidator {
 
+	public String getName();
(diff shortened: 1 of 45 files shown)


common/client/src/main/java/zingg/common/client/license/ILicenseValidator.java
```diff
@@ -0,0 +1,8 @@
+package zingg.common.client.license;
+
+public interface ILicenseValidator {
+
+	
+	public boolean validate();
```
sonalgoyal, line 6: will probably need a name by which it will be looked up in the IZinggLicense

common/client/src/main/java/zingg/common/client/license/IZinggLicense.java
```diff
@@ -0,0 +1,8 @@
+package zingg.common.client.license;
+
+public interface IZinggLicense {
```
sonalgoyal, line 3: its not clear to me how this will evolve to reading, loading and passing the license constraints to the validators. If that's clear to you, please ignore and we can discuss during the implementation of the concrete classes.

common/client/src/main/java/zingg/common/client/license/ILicenseValidator.java
vikasgupta78: added getName method in commit e9a57412

common/client/src/main/java/zingg/common/client/license/IZinggLicense.java
vikasgupta78: let's discuss during the implementation of the concrete classes

PR #628, opened 2023-07-31
Obvious dupes performance optimisation

vikasgupta78:
Obvious dupes performance optimisation


sonalgoyal (changes requested):
looks good to me but it will be great if you can bring back the commented code - its used for quick debugs when things are going wrong,

Issue #284, opened 2022-05-24
Add AGPL license header to all code. 


sonalgoyal:

vishalshetty104:
Please assign this to me

PR #638, opened 2023-08-08
AGPL License 

vishalshetty104:
##Reference Issue: #284 

Added AGPL license information to all source code files: .Java and .Py in the project

### DIFF
--- a/common/client/src/main/java/zingg/common/client/Arguments.java
+++ b/common/client/src/main/java/zingg/common/client/Arguments.java
@@ -1,3 +1,22 @@
+/*
+ * Zingg
+ * Copyright (C) 2021-Present  Zingg Labs,inc
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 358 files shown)


sonalgoyal:
@vikasgupta78 please review

vikasgupta78:
@vishalshetty104 please share the script used in making these changes

vishalshetty104:
@vikasgupta78 i used Intelli-J 's built-in copyright generator.
I used this guide - https://www.jetbrains.com/help/idea/copyright.html

vikasgupta78:
This would mean any new files in future would depend on the developer whether developer adds it or not.

It would be better to write a script which when run would:

1. ask which all files e.g. dir to start from and file pattern to check
2. checks that file has a header or not , if not puts it
3. Header is stored at a common location so that can be updated
=> may need different due to way comments work for java vs py etc.
5. Has option to delete existing header and put new if required.

Current changes may work in short term but difficult to maintain / review specially since we have open source contributors.

Shell script is a good way to go about it but I am open for suggestions.

vikasgupta78:
If there is a way for intellij to take care of this, I am ok with no script also

vishalshetty104:
I initially planned on doing a .sh script but it was not working. The Intelli-j option seemed like an alternative solution, hence i went for it.

Once u set up the copyright profile like I did, Intelli-j automatically inserts the copyright license whenever a new file is created. If incase we need to update the license information (Eg: change year) then we just need to go into project settings and edit the copyright profile. 

Once thats done, we just need to right click on the project root node and click on update copyright.
For me, this has worked for both .java as well as .py files

vishalshetty104:
But this solution does limit to only Intelli-J users, I will try to get the shell script to work and share it here.

PR #647, opened 2023-08-17
Instructions on prerequisite software needed to compile / run zingg

vikasgupta78:


sonalgoyal (changes requested):
These changes should be part of setting up the dev env. We should leave the user guide as is.

vikasgupta78:
docs/settingUpZingg.md updated as per discussion in morning prerequisites.md removed

PR #683, opened 2023-10-06
UTs for clients.py

gnanaprakash-ravi:
Unittests for some clients.py functions.


vikasgupta78:
can you also git mv the two defunct test files e.g.:

git mv testInfra.py defucnt-TestInfra.py


after that test they are not getting triggered and commit

gnanaprakash-ravi:
ok sir!!

Issue #684, opened 2023-10-06
Broken links in Zingg docs


sonalgoyal:
_configurations_ on: <https://docs.zingg.ai/zingg/stepbystep/installation/docker/shared-locations>

_Zingg config_ on: <https://www.zingg.ai/documentation-article/identifying-duplicates-in-snowflake-with-zingg>

[Slack Message](https://zinggai.slack.com/archives/C02JNH144TB/p1696610627606969?thread_ts=1696547868.614449&cid=C02JNH144TB)

sonalgoyal:
@gnanaprakash-ravi

PR #685, opened 2023-10-09
Fix broken docs links

gnanaprakash-ravi:
This PR is for the issue [684](https://github.com/zinggAI/zingg/issues/684)


sonalgoyal (changes requested):
Can oyu please change the paths to relative paths?

gnanaprakash-ravi:
ok ma'am

PR #689, opened 2023-10-14
Change model fields PHASE

vikasgupta78:
Change model fields PHASE


vikasgupta78:
This is now out of sync with 0.4.0

Issue #692, opened 2023-10-17
ObvDupe filter for ftd


vikasgupta78:
1. we dont want users to label what they already specify as obvious dupes

2. hypothetically if say all 100% records are filtered due to obv dupe

      what would ftd do?
      
      It would do nothing. print a message

vikasgupta78:
Also move obvDupe to common class to be used by matcher and ftd

vikasgupta78:
Possible Impact on:

1. ZFrame
2. SparkFrame
3. TrainingDataFinder
4. Matcher

vikasgupta78:
scenarios to be tested for ftd:

1. No obv dupe [but obv dupe condition there]
2. Everything obv dupe i.e. no ftd left
3. Some obv dupe
4. No obv dupe condition there

vikasgupta78:
For scenario 2 i.e. Everything obv dupe i.e. no ftd left :

It worked fine, no unmarked data created and following message when label run:

It seems there are no unmarked records at this moment. Please run findTrainingData job to build some pairs to be labelled and then run this labeler.

sonalgoyal:
it should suggest to the user that it is unable to find any pairs as all pairs it sampled are part of the obvious duplicate condition

vikasgupta78:
We should do this in ftd stage?

vikasgupta78:
Added the message in TrainingDataFinder

vikasgupta78:
There is no impact on linker as linker is just extending from Matcher and all obv dupe related work is being handled by Matcher

vikasgupta78:
PR #694 Merged

PR #694, opened 2023-10-18
Issue #692 obv dupe changes for ftd and matcher

vikasgupta78:
Issue #692 obv dupe changes for ftd and matcher


common/core/src/main/java/zingg/common/core/executor/ObvDupeFilter.java
```diff
@@ -0,0 +1,247 @@
+package zingg.common.core.executor;
```
sonalgoyal, line 1: this class is not an executor and should be broken down into two classes at least. they can live in a separate package zingg.common.core.obviousDupes. one that takes args, dsutil and creates filters. one that applies those.

common/core/src/main/java/zingg/common/core/executor/Matcher.java
```diff
@@ -126,7 +121,7 @@ public void execute() throws ZinggClientException {
 			ZFrame<D, R, C> obvDupePairs = getObvDupePairs(blocked);
```
sonalgoyal, line 121: this logic should move to obvious dupes. the matcher should just ask the obv dupes to add/remove dupes as necessary by passing blocks/pairs whatever

common/core/src/main/java/zingg/common/core/executor/ObvDupeFilter.java
```diff
@@ -0,0 +1,247 @@
...
+	
+	/**
+	 * 
+	 * obviousDupeString format col1 & col2 | col3 | col4 & col5
+	 * 
+	 * @param obviousDupeString
+	 * @return
+	 */
+	
+	public C getObviousDupesFilter(ZFrame<D, R, C> df1, ZFrame<D, R, C> dfToJoin, String obviousDupeString, C extraAndCond) {
```
sonalgoyal, line 139: if you define the obvious filter not through a string put through json, this code becomes much simpler. how about 

"obviousDupes": [
"matchCondition": [{"column":"fname"},{"column":"lname"}]
"matchCondition":[{}]
]

You can simply define this object as part of the args. It gets populated automatically, no need to parse, trim etc.

sonalgoyal (changes requested):

common/core/src/main/java/zingg/common/core/executor/Matcher.java
vikasgupta78: The logic is already moved , this method delegates work to the obv dupe class. The method is kept as it gets overriden in enterprise matcher for purpose of instrumentation

vikasgupta78: That is getObvDupePairs method in Matcher

common/core/src/main/java/zingg/common/core/executor/ObvDupeFilter.java
vikasgupta78: Will create a JSON sample for various scenarios and get it reviewed before implementation

common/core/src/main/java/zingg/common/core/executor/ObvDupeFilter.java
vikasgupta78: I had thought that may be we might want to create a phase out of it in future? 100% match filter kinds? If no such plan can break it down, no problem

sonalgoyal: So we are talking of two different things here. 
- A new possible phase for obviousDupes. If we do that, we need some kind of executor. 
- Breaking down the class 

The first one can always be defined later using the classes in second.

common/core/src/main/java/zingg/common/core/executor/Matcher.java
sonalgoyal: then we should override the new class in enterprise and introduce im and weave into the ent matcher. 

Obvious dupes logic should be encapsulated in its own package and classes.

sonalgoyal: I went back and checked the code. The following should not be part of the Matcher. 
`if (obvDupePairs != null) {
				long obvDupeCount = obvDupePairs.count();
				LOG.debug("obvDupePairs count " + obvDupeCount);
				if (obvDupeCount > 0) {
					blocks = removeObvDupesFromBlocks(blocks);
				}
			}
			`

The methods in the Matcher that you have defined can continue to call the obv dupes classes.

vikasgupta78: done in commit 70f1c33e, please review

common/core/src/main/java/zingg/common/core/executor/ObvDupeFilter.java
vikasgupta78: Done in commit c2ddce9d please review

common/core/src/main/java/zingg/common/core/executor/ObvDupeFilter.java
vikasgupta78: Refactor to two seperate classes in commit a3d77a68 and f3480e15

vikasgupta78:
have completed all review comments of obv dupe filter in OSS
once you have merged than will do enterprise side of changes

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeFilter.java
```diff
@@ -0,0 +1,126 @@
...
+import zingg.common.client.util.ColName;
+import zingg.common.core.Context;
+
+public class ObvDupeFilter<S,D,R,C,T> implements Serializable {
+	
+	private static final long serialVersionUID = 1L;
+	public static final Log LOG = LogFactory.getLog(ObvDupeFilter.class); 
+
+    protected Arguments args;	
+    protected Context<S,D,R,C,T> context;
```
sonalgoyal, line 20: Do we need to pass Context or can we pass DSUtil ?

sonalgoyal: also see if S and T are needed for this class?

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeFilter.java
```diff
@@ -0,0 +1,126 @@
...
+			return null;
+		}
+
+		ZFrame<D,R,C> prefixBlocked = context.getDSUtil().getPrefixedColumnsDS(blocked);		
+		C gtCond = blocked.gt(prefixBlocked,ColName.ID_COL);
+		
+		ZFrame<D,R,C> onlyIds = null;
+		
+		// loop thru the values and build a filter condition		
+		for (int i = 0; i < obviousDupes.length; i++) {
```
sonalgoyal, line 63: is tyhere a reason to loop through this one by one? I thought the filter would return the and, or conditions to join on completely?

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeFilter.java
```diff
@@ -0,0 +1,126 @@
...
+	 */
+	public ZFrame<D, R, C> removeObvDupesFromBlocks(ZFrame<D, R, C> blocks) {
+		
+		LOG.debug("blocks count before removing obvDupePairs " + blocks.count());
+		ObviousDupes[] obviousDupes = args.getObviousDupes();
+		if (obviousDupes == null || obviousDupes.length == 0) {
+			return blocks;
+		}
+		C reverseOBVDupeDFFilter = obvDupeUtil.getReverseObviousDupesFilter(blocks,obviousDupes,null);
+		if (reverseOBVDupeDFFilter != null) {
```
sonalgoyal, line 108: can our zframe methods take care of this so that we dont have to do null checks in main code?

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeUtil.java
```diff
@@ -0,0 +1,190 @@
...
+	public C getReverseObviousDupesFilter(ZFrame<D, R, C> df1,ZFrame<D,R,C> dfToJoin, ObviousDupes[] obviousDupes, C extraAndCond) {
+		return df1.not(getObviousDupesFilter(df1,dfToJoin,obviousDupes,extraAndCond));
+	}
+	
+	/**
+	 * Add prediction and score cols
+	 * @param obvDupes
+	 * @return
+	 */
+	public ZFrame<D,R,C> massageObvDupes(ZFrame<D,R,C> obvDupes) {
```
sonalgoyal, line 184: this belongs to the other class. It seems that the class names can be swapped - ObvDupeUtil can be renamed to ObviousDupeCondiiton or even ObvDupeFilter as it is only building the join/filter conditions.

sonalgoyal (changes requested):

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeFilter.java
vikasgupta78: We have to loop thru all the match conditions and within match conditions for every field name.

e.g "obviousDupes":[{   "matchCondition":[  { "fieldName":"fname"  },  { "fieldName":"stNo"  },  { "fieldName":"add1"  }   ]},{"matchCondition":[   {  "fieldName":"ssn"   }] }, {   "matchCondition":[{"fieldName":"fname" }, {"fieldName":"stNo" }, {"fieldName":"lname" }  ]} ],


so we loop thru to get 3 matchCondition and then within each matchCondition for each fieldName

forming something like 
(fname = z_fname and stNo = z_stNo and add1 = z_add1) OR 
(ssn = z_ssn) OR
(fname = z_fname and stNo = z_stNo and lname = z_lname)

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeUtil.java
vikasgupta78: Please tell the names of both and I will do it.

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeFilter.java
vikasgupta78: As Column can be used in various contexts => null check at all places where colum is used.

This is better

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeFilter.java
vikasgupta78: done in commit b29758c5 , please review

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeUtil.java
vikasgupta78: method moved : commit 45e9205a

sonalgoyal: ObvDupeUtil is the class that applies the filters, adds and removes obv dupes 
ObvDupeCondition or ObvDupeFilter is the class that reads the args vaues and creates the conditions

sonalgoyal (changes requested):

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeUtil.java
vikasgupta78: Renamed in commit 6db0092d , please review

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeFilter.java
vikasgupta78: you are right as it will come in play only if obvious dupes there , removed redundant null in commit eb6b42f6

please review

common/core/src/main/java/zingg/common/core/obviousdupes/ObvDupeFilter.java
vikasgupta78: This is query formation.

We don't actually use it exactly like this.

In case of getting obv dupes from blocked:

1. We use each and condition
2. union all the dfs got from above
3. do distinct
Directly doing OR was causing performance issues

For removing obv dupe from blocks
We first form a combined condition like above and than return a combined condition with NOT:
NOT (
(fname = z_fname and stNo = z_stNo and add1 = z_add1) OR
(ssn = z_ssn) OR
(fname = z_fname and stNo = z_stNo and lname = z_lname)
)

The way we make Column filter expression is independent of how we use it.

vikasgupta78: Basically we don;t use one final Column in end and we are flexible. You have asked similar question below, I hope this answers both

vikasgupta78: By keeping build of column of filter expression independent of how we use it we can use it as needed

vikasgupta78: also many of the methods are private only the public/protected can be used by outside code

Issue #709, opened 2023-10-27
config json in examples has wrong order in schema


vikasgupta78:
"schema": "recId string, fname string, lname string, stNo string, add1 string, add2 string, city string, state string, areacode string, dob string, ssn  string" 

it should be :

"schema": "recId string, fname string, lname string, stNo string, add1 string, add2 string, city string, areacode string, state string,  dob string, ssn  string" 

That is as per the data in test.csv areacode should come before state

May have similar bug in other configs

vikasgupta78:
@gnanaprakash-ravi  can you please look into this?

gnanaprakash-ravi:
Ok sir, I will look into it

PR #711, opened 2023-10-27
corrected order in schema

gnanaprakash-ravi:
This PR is for the issue [709](https://github.com/zinggAI/zingg/issues/709)


vikasgupta78:
What all testing have you done? And are these the only 2 configs?

vikasgupta78:
Also is there an impact on FebrlExample.py

vikasgupta78:
When I say impact on FebrlExample.py I mean order needs to be same in both as both use test.csv

gnanaprakash-ravi:
Yes FebrlExample.py also, areacode was after the state, and i changed it, i will run the 4 phases on config.json and FebrlExample.py testing!!

vikasgupta78:
@gnanaprakash-ravi after the merge of PR for det match changes , you might want to revisit this as this may give conflict in merge

Issue #730, opened 2023-11-28
Business Exception


vikasgupta78:
To handle business scenarios e.g. empty data pipe we should have Business Exception for graceful exit

PR #731, opened 2023-11-28
added business exception class, issue #730

vikasgupta78:


common/client/src/main/java/zingg/common/client/ZinggBusinessException.java
```diff
@@ -0,0 +1,21 @@
+package zingg.common.client;
+
+import java.io.Serializable;
+
+/**
+ * To be thrown in case of business scenario which needs graceful handling
+ *
+ */
+public class ZinggBusinessException extends Exception implements Serializable {
```
sonalgoyal, line 9: Do we need to extend Serializable given Exception already extends Serializable?

common/client/src/main/java/zingg/common/client/ZinggBusinessException.java
```diff
@@ -0,0 +1,21 @@
...
+
+	private static final long serialVersionUID = 1L;
+	
+	public ZinggBusinessException(String message) {
+		super(message);
+	}
+
+	public ZinggBusinessException(String message, Throwable cause) {
```
sonalgoyal, line 17: should this be a throwable or an exception?

sonalgoyal (changes requested):

common/client/src/main/java/zingg/common/client/ZinggBusinessException.java
vikasgupta78: Throwable => Error as well => no graceful handling, no alternate route, no recovery scenarios

We want graceful handling so we need Exception as root not Throwable as it would never be Error

common/client/src/main/java/zingg/common/client/ZinggBusinessException.java
vikasgupta78: I did it because that's how it is in ZinggClientException

common/client/src/main/java/zingg/common/client/ZinggBusinessException.java
sonalgoyal: Are you talking about line 17 here or the hierarchy?

vikasgupta78: hierarchy

vikasgupta78: The constructors are based on constructors of exception class

https://docs.oracle.com/javase/8/docs/api/java/lang/Exception.html

sonalgoyal: ok, makes sense.

common/client/src/main/java/zingg/common/client/ZinggBusinessException.java
vikasgupta78: ya I can fix it. I thought it's done in ZinggClientException so that's how we should be doing it.

sonalgoyal: no, feel free to point out issues, log bugs and fix whatever you can.

vikasgupta78: fixed in commit dc794dba

Issue #723, opened 2023-11-17
show concise generate docs


vikasgupta78:
show concise option in generate docs is needed, so that dont use columns do not come.

PR #750, opened 2023-12-27
generate docs to show concise data if user so asks issue #723

vikasgupta78:
generate docs to show concise data if user so asks issue #723

### DIFF
--- a/common/client/src/main/java/zingg/common/client/Arguments.java
+++ b/common/client/src/main/java/zingg/common/client/Arguments.java
@@ -171,20 +171,6 @@ public class Arguments implements Serializable, IArguments {
 		return fieldDefinition;
 	}
 	
-	@JsonIgnore @Override
-	public List<? extends FieldDefinition> getFieldDefinitionDontUse() {
-		return getFieldDefinition().stream()
-			    .filter(x->x.matchType.contains(MatchType.DONT_USE))
-			    .collect(Collectors.toList());
-	}
-	
-	@JsonIgnore @Override
-	public List<? extends FieldDefinition> getFieldDefinitionToUse() {
-		return getFieldDefinition().stream()
-			    .filter(x->!x.matchType.contains(MatchType.DONT_USE))
-			    .collect(Collectors.toList());
-	}
-	
 	/**
 	 * Set the field definitions consisting of match field indices, types and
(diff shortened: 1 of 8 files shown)


common/client/src/main/java/zingg/common/client/Arguments.java
```diff
@@ -169,7 +170,21 @@ public void setLabelDataSampleSize(float labelDataSampleSize) throws ZinggClient
 	public List<? extends FieldDefinition> getFieldDefinition() {
 		return fieldDefinition;
 	}
-
+	
+	@JsonIgnore @Override
```
sonalgoyal, line 174: we should move this to a separate class. args dont need to know about match types

sonalgoyal (changes requested):

common/client/src/main/java/zingg/common/client/Arguments.java
vikasgupta78: What would be a good place? I could have kept it in ModelDocumenter but keeping it at common place might be useful to others in future.

sonalgoyal: yes it is likely useful to others. I think a FieldDefUtil may be ok for now.

vikasgupta78: Done in latest commit

common/client/src/test/java/zingg/common/client/TestArguments.java
```diff
@@ -25,6 +26,8 @@ public class TestArguments {
 
 	public static final Log LOG = LogFactory.getLog(TestArguments.class);
 	protected ArgumentsUtil argsUtil = new ArgumentsUtil();
+	
+	protected FieldDefUtil fieldDefUtil = new FieldDefUtil();
```
sonalgoyal, line 30: These tests should be moved to a separate class for the util.

sonalgoyal (changes requested):
Junit changes

common/client/src/test/java/zingg/common/client/TestArguments.java
vikasgupta78: Done in latest commit

PR #814, opened 2024-04-08
[WIP][DO NOT MERGE] Draft implementation of the new PySpark API for support of both Spark Classic and Spark Connect

SemyonSinchenko:
The main idea is to make python part lazy. All the classes, like Argument, Pipe, etc. should be pure python classes that do not contain any interactions with JVM.

Interaction with JVM in this case should be encapsulated into 2-3 methods of the new `Client` and will contain passing the full JSON generated from python Arguments, Pipes, FieldDefintions, etc.


python/zingg_v2/models.py
```diff
@@ -0,0 +1,133 @@
...
+    PINCODE = auto()
+    NULL_OR_BLANK = auto()
+    TEXT = auto()
+    NUMERIC = auto()
+    NUMERIC_WITH_UNITS = auto()
+    ONLY_ALPHABETS_EXACT = auto()
+    ONLY_ALPHABETS_FUZZY = auto()
+
+
+class FileFormat(StrEnum):
```
sonalgoyal, line 25: lets rename to DataFormat or plain format?

python/zingg_v2/models.py
```diff
@@ -0,0 +1,133 @@
...
+    JDBC = auto()
+    CASSANDRA = "org.apache.spark.sql.cassandra"
+    SNOWFLAKE = "net.snowflake.spark.snowflake"
+    ELASTIC = "org.elasticsearch.spark.sql"
+    EXACOL = "com.exasol.spark"
+    BIGQUERY = auto()
+    INMEMORY = auto()
+
+
+class FieldPreprocessor(StrEnum):
```
sonalgoyal, line 41: this is not used in the code and doe snot need to be exposed

python/zingg_v2/models.py
```diff
@@ -0,0 +1,133 @@
...
+    stopWords: Optional[str] = None
+    abbreviations: Optional[str] = None
+
+
+class Pipe(BaseModel):
+    name: str
+    format: FileFormat
+    preprocessors: Optional[FieldPreprocessor] = None
+    props: dict[str, Any] = {}
+    schema: Optional[str] = None
```
sonalgoyal, line 74: only a csv pipe has the schema

python/zingg_v2/models.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class Arguments(BaseModel):
+    output: Optional[list[Pipe]] = None
+    data: Optional[list[Pipe]] = None
+    zinggDir: str = "/tmp/zingg"
+    trainingSamples: Optional[list[Pipe]] = None
+    fieldDefinition: Optional[list[FieldDefinition]] = None
+    numPartitions: int = 10
+    labelDataSampleSize: float = 0.01
```
sonalgoyal, line 85: how do we ensure that stuff we define in java remains the same here?

python/zingg_v2/models.py
SemyonSinchenko: Yes, that is why the default value is `None`. If I understand correctly the Java part, even if I pass the schema with format equal `parquet` this schema will be just ignored.

python/zingg_v2/models.py
SemyonSinchenko: Deleted.

python/zingg_v2/models.py
SemyonSinchenko: Done.

PR #815, opened 2024-04-09
using data selector as wrapper over raw data

vikasgupta78:
using data selector as wrapper over raw data

### DIFF
--- a/common/core/src/main/java/zingg/common/core/data/df/ZFrameDataSelector.java
+++ b/common/core/src/main/java/zingg/common/core/data/df/ZFrameDataSelector.java
@@ -1,0 +1,75 @@
+package zingg.common.core.data.df;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import zingg.common.client.IArguments;
+import zingg.common.client.ZFrame;
+import zingg.common.client.ZinggClientException;
+import zingg.common.client.cols.ZidAndFieldDefSelector;
+import zingg.common.client.util.ColName;
+import zingg.common.core.block.Canopy;
+import zingg.common.core.block.Tree;
+import zingg.common.core.context.Context;
+import zingg.common.core.preprocess.StopWordsRemover;
+import zingg.common.core.util.Analytics;
+import zingg.common.core.util.Metric;
+
+public class ZFrameDataSelector<S, D, R, C, T> {
+
+	protected ZFrame<D,R,C> rawData;
+	protected ZFrame<D,R,C> fieldDefColumnsDS;
+	protected ZFrame<D,R,C> blockedData;
+	protected IArguments args;
+	protected Context<S,D,R,C,T> context;
+	protected StopWordsRemover<S, D, R, C, T> stopWordsRemover;
+	protected ZFrame<D, R, C> preprocessedRepartitionedData;
+	
+	public static final Log LOG = LogFactory.getLog(ZFrameDataSelector.class);   
+	
+	public ZFrameDataSelector(ZFrame<D, R, C> rawData, IArguments args, Context<S,D,R,C,T> context,StopWordsRemover<S, D, R, C, T> stopWordsRemover) {
+		super();
+		this.rawData = rawData;
+		this.args = args;
+		this.context = context;
+		this.stopWordsRemover = stopWordsRemover;
+	}
+
+	public ZFrame<D, R, C> getRawData() {
+		return rawData;
+	}
+
+	public ZFrame<D, R, C> getFieldDefColumnsDS() {
+		if (fieldDefColumnsDS==null) {
+			ZidAndFieldDefSelector zidAndFieldDefSelector = new ZidAndFieldDefSelector(args.getFieldDefinition());
+			fieldDefColumnsDS = getRawData().select(zidAndFieldDefSelector.getCols());
+		}
+		return fieldDefColumnsDS;
+//		return getDSUtil().getFieldDefColumnsDS(testDataOriginal, args, true);
+	}
+	
+	public ZFrame<D, R, C> getPreprocessedRepartitionedData() {
+		return preprocessedRepartitionedData;
+	}
+
+	public ZFrame<D,R,C>  getBlocked() throws Exception, ZinggClientException{
+		if (blockedData==null) {
+			preprocessedRepartitionedData = stopWordsRemover.preprocessForStopWords(getFieldDefColumnsDS());
+			preprocessedRepartitionedData = preprocessedRepartitionedData.repartition(args.getNumPartitions(), preprocessedRepartitionedData.col(ColName.ID_COL));
+			//testData = dropDuplicates(testData);
+			long count = preprocessedRepartitionedData.count();
+			LOG.info("Read " + count);
+			Analytics.track(Metric.DATA_COUNT, count, args.getCollectMetrics());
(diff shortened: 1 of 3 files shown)


common/core/src/main/java/zingg/common/core/data/df/ZFrameDataSelector.java
```diff
@@ -0,0 +1,75 @@
+package zingg.common.core.data.df;
```
sonalgoyal, line 1: rename to ZData

common/core/src/main/java/zingg/common/core/data/df/ZFrameDataSelector.java
```diff
@@ -0,0 +1,75 @@
...
+	protected ZFrame<D,R,C> fieldDefColumnsDS;
+	protected ZFrame<D,R,C> blockedData;
+	protected IArguments args;
+	protected Context<S,D,R,C,T> context;
+	protected StopWordsRemover<S, D, R, C, T> stopWordsRemover;
+	protected ZFrame<D, R, C> preprocessedRepartitionedData;
+	
+	public static final Log LOG = LogFactory.getLog(ZFrameDataSelector.class);   
+	
+	public ZFrameDataSelector(ZFrame<D, R, C> rawData, IArguments args, Context<S,D,R,C,T> context,StopWordsRemover<S, D, R, C, T> stopWordsRemover) {
```
sonalgoyal, line 30: this class shuld not be aware of the blocking tree loading and context etc. it should get the blocking tree and have a method to apply it.

common/core/src/main/java/zingg/common/core/data/df/ZFrameDataSelector.java
```diff
@@ -0,0 +1,75 @@
...
+//		return getDSUtil().getFieldDefColumnsDS(testDataOriginal, args, true);
+	}
+	
+	public ZFrame<D, R, C> getPreprocessedRepartitionedData() {
+		return preprocessedRepartitionedData;
+	}
+
+	public ZFrame<D,R,C>  getBlocked() throws Exception, ZinggClientException{
+		if (blockedData==null) {
+			preprocessedRepartitionedData = stopWordsRemover.preprocessForStopWords(getFieldDefColumnsDS());
```
sonalgoyal, line 57: should be a separate method

common/core/src/main/java/zingg/common/core/data/df/ZFrameDataSelector.java
```diff
@@ -0,0 +1,75 @@
...
+			preprocessedRepartitionedData = preprocessedRepartitionedData.repartition(args.getNumPartitions(), preprocessedRepartitionedData.col(ColName.ID_COL));
+			//testData = dropDuplicates(testData);
+			long count = preprocessedRepartitionedData.count();
+			LOG.info("Read " + count);
+			Analytics.track(Metric.DATA_COUNT, count, args.getCollectMetrics());
```
sonalgoyal, line 62: count should be a method on this and Analytics should not be a part of this object - it should be driven from the executor

sonalgoyal (changes requested):

common/core/src/main/java/zingg/common/core/data/df/IZFrameProcessor.java
```diff
@@ -0,0 +1,11 @@
+package zingg.common.core.data.df;
+
+import zingg.common.client.ZFrame;
+
+public interface IZFrameProcessor<S, D, R, C, T> {
```
sonalgoyal, line 5: does it need S and T?

common/core/src/main/java/zingg/common/core/data/df/IZFrameProcessor.java
```diff
@@ -0,0 +1,11 @@
+package zingg.common.core.data.df;
```
sonalgoyal, line 1: it may be better to introduce a method process() so that the logic is not in the constructor.

common/core/src/main/java/zingg/common/core/data/df/FieldDefFrame.java
```diff
@@ -0,0 +1,38 @@
...
+	
+	protected IArguments args;
+	
+	public static final Log LOG = LogFactory.getLog(FieldDefFrame.class);   
+	
+	public FieldDefFrame(ZFrame<D, R, C> originalDF, IArguments args) {
+		super();
+		this.originalDF = originalDF;
+		this.args = args;
+		this.processedDF = getOriginalDF().select(new ZidAndFieldDefSelector(args.getFieldDefinition()).getCols());
```
sonalgoyal, line 24: better to have an overloaded constructor here which can take the fielddefselectors. we have a few so may need to pass those in the future

common/core/src/main/java/zingg/common/core/data/df/RepartitionFrame.java
```diff
@@ -0,0 +1,37 @@
...
+
+	protected ZFrame<D,R,C> originalDF;
+	
+	protected ZFrame<D,R,C> processedDF;
+	
+	protected IArguments args;
+	
+	public static final Log LOG = LogFactory.getLog(RepartitionFrame.class);   
+	
+	public RepartitionFrame(ZFrame<D, R, C> originalDF, IArguments args) {
```
sonalgoyal, line 20: we should pass the numberofPartitions and column on which we need to partition

common/core/src/main/java/zingg/common/core/data/df/PreprocessedFrame.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import zingg.common.client.IArguments;
+import zingg.common.client.ZFrame;
+import zingg.common.client.ZinggClientException;
+import zingg.common.core.preprocess.StopWordsRemover;
+
+public class PreprocessedFrame<S, D, R, C, T> implements IZFrameProcessor<S, D, R, C, T> {
```
sonalgoyal, line 11: change stopword design - have a processor there. so preprocessedframe can be passed any processor(stopwords, tomorrow lowercase etc)

sonalgoyal (changes requested):
I have added few comments. Broadly these classes should not know about args, context etc. BlockedFrame can be passed the blockingtreeutil. We need to be able to test these classes and setup for that has to be smooth/mockable

common/core/src/main/java/zingg/common/core/data/df/IZFrameProcessor.java
vikasgupta78: BlockingTreeUtil<S, D,R,C,T>

vikasgupta78: StopWordsRemover<S,D,R,C,T>

sonalgoyal: so we should have one frame<D,R,C> and then another that extends this and adds S and T.

vikasgupta78: what benefit it would give? ZData in any case has <S,D,R,C,T>

vikasgupta78: unnecessarily two very similar interfaces to be maintained whose objects can't be assigned to one another

Issue #821, opened 2024-04-18
Support for other feature types in non string fields


vikasgupta78:
Support for other feature types in non string fields

Currently only fuzzy is there

sonalgoyal:
we can add null_or_blank to integer, long and dates. other match types are not applicable. please also update docs https://docs.zingg.ai/zingg0.4.0/stepbystep/configuration/field-definitions

vikasgupta78:
As discussed exact might have to be supported for numeric types

vikasgupta78:
added null or blank and exact match type to int, long and date pr #824

sonalgoyal:
@vikasgupta78 ok to close?

PR #824, opened 2024-04-23
added null or blank and exact match type to int, long and date issue#821

vikasgupta78:
added null or blank and exact match type to int, long and date issue #821


sonalgoyal (changes requested):
we should be able to templatize the checknullfunction as well as the similarityfucnton so it can be used with any data type

vikasgupta78:
> we should be able to templatize the checknullfunction as well as the similarityfucnton so it can be used with any data type

done in commit ba3503f5

PR #849, opened 2024-07-02
Update match.md

Nitish1814:
view match results in console

### DIFF
--- a/docs/setup/match.md
+++ b/docs/setup/match.md
@@ -14,7 +14,7 @@ As can be seen in the image below, matching records are given the same z_cluster
 
 ![Match results](/assets/match.gif)
 
-The match results in CSV format will be saved in /tmp/zinggOutput.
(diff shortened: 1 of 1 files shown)


docs/setup/match.md
```diff
@@ -14,4 +14,17 @@ As can be seen in the image below, matching records are given the same z_cluster
 
 ![Match results](/assets/match.gif)
 
-If records across multiple sources have to be matched, the [link phase](./link.md) should be used.
\ No newline at end of file
+The match results in CSV format will be saved in /tmp/zinggOutput.
```
sonalgoyal, line 17: The match results are saved in zinggDir/modelId or in the respective database, pipe, NoSQL db etc. They may not be csv either. So lets skip that. 

We can try to actually create an InMemoryPipe from the output pipe we defined and read that. See if you can come up with something here.

PR #850, opened 2024-07-03
Doc refactor

Nitish1814:
Refactored setup doc

### DIFF
--- a/docs/settingUpZingg.md
+++ b/docs/settingUpZingg.md
@@ -92,9 +92,11 @@ vim ~/.bashrc
 export SPARK_HOME=/opt/spark
 export SPARK_MASTER=local[\*]
 export MAVEN_HOME=/home/ubuntu/apache-maven-3.8.8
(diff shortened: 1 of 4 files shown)


docs/settingUpZingg.md
```diff
@@ -35,13 +35,13 @@ _**Note :-**_ It is suggested to fork the repository to your account and then cl
 
 ****
 
-_**Step 2 :  Install JDK 1.8 (Java Development Kit)**_
```
sonalgoyal, line 38: the jdk change has to go into the installation page as well. can you please change, verify and update this PR?

sonalgoyal: also please check the docker image and which jdk it has. jdk 11 was done only last week and we havent fully decided if we are going with it or not.

sonalgoyal (changes requested):

docs/settingUpZingg.md
Nitish1814: done, docker image is already on JAVA 11

PR #859, opened 2024-07-10
Block and stop word test

Nitish1814:
Block and StopWord Tests


vikasgupta78:
as discussed I am not convinced that we need IWithSession and WithSession in the main code. Also naming needs to be fixed. Do an impact analysis and discuss with @sonalgoyal

vikasgupta78:
there was already a class TestBlock in zingg spark core? Why is that not being reused?

vikasgupta78:
there was existing TestStopWords class why it has been deleted instead of renamed / moved to maintain history

Nitish1814:
> there was already a class TestBlock in zingg spark core? Why is that not being reused?

just restructured package, reused same file only !

Nitish1814:
> there was existing TestStopWords class why it has been deleted instead of renamed / moved to maintain history

1: TestStopWords first made generic
2: Renamed to TestStopWordsBase
3: Moved to common

Nitish1814:
> there was already a class TestBlock in zingg spark core? Why is that not being reused?

1: TestBlock first made generic
2: Renamed to TestBlockBase
3: Moved to common

Nitish1814:
> PojoToArrayConverter

removed

PR #862, opened 2024-07-11
fix for multiple simpleFunction register

Nitish1814:
Multiple simpleFunction registery fix done!

Needed to register udf only if it is not already registered

### DIFF
--- a/spark/core/src/main/java/zingg/spark/core/model/VectorValueExtractor.java
+++ b/spark/core/src/main/java/zingg/spark/core/model/VectorValueExtractor.java
@@ -9,6 +9,7 @@ import org.apache.spark.sql.types.DataTypes;
 
 import org.apache.spark.sql.SparkSession;
 import zingg.spark.core.similarity.SparkBaseTransformer;
+import zingg.spark.core.util.SparkFnRegistrar;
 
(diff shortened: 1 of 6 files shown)


spark/core/src/main/java/zingg/spark/core/model/VectorValueExtractor.java
```diff
@@ -26,7 +26,11 @@ public Double call(Vector v) {
 	
 	@Override
 	public void register(SparkSession spark) {
-    	spark.udf().register(uid, (UDF1) this, DataTypes.DoubleType);
+
+		//only register udf if it is not registered already
+		if (!spark.catalog().functionExists(uid)) {
```
sonalgoyal, line 31: please create common class SparkFnRegistrar which takes SparkSession and fn params, does the check and registers accordingly.

sonalgoyal (changes requested):
avoid if then else in main code,
move common checks to single place

spark/core/src/main/java/zingg/spark/core/model/VectorValueExtractor.java
Nitish1814: done

PR #923, opened 2024-10-18
VerifyBlocking Phase Draft PR

sania-16:


common/core/src/main/java/zingg/common/core/block/Blocker.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public ZFrame<D,R,C>  getBlocked(ZFrame<D,R,C> testData) throws Exception, ZinggClientException{
+		LOG.debug("Blocking model file location is " + args.getBlockFile());
+		Tree<Canopy<R>> tree = getBlockingTreeUtil().readBlockingTree(args);
+		ZFrame<D,R,C> blocked = getBlockingTreeUtil().getBlockHashes(testData, tree);		
+		ZFrame<D,R,C> blocked1 = blocked.repartition(args.getNumPartitions(), blocked.col(ColName.HASH_COL)).cache();
+		return blocked1;
+	}
+
+    public Pipe<D,R,C> getPipeForDebugBlockingLocation(long timestamp, String type){
```
sonalgoyal, line 38: rename to verify instead of debug

common/core/src/main/java/zingg/common/core/block/Blocker.java
```diff
@@ -0,0 +1,93 @@
...
+			testDataOriginal =  getFieldDefColumnsDS(testDataOriginal).cache();
+			ZFrame<D,R,C> blocked = getBlocked(testDataOriginal);
+			LOG.info("Blocked");
+			setTimestamp(timestamp);
+
+            getPipeUtil().write(blocked.select(ColName.HASH_COL).groupByCount(ColName.HASH_COL, ColName.HASH_COL + "_count").sortDescending(ColName.HASH_COL + "_count"), getPipeForDebugBlockingLocation(timestamp, "counts"));	
+
+            blocked.limit(3);
+
+            getPipeUtil().write(blocked.sample(false, 0.1), getPipeForDebugBlockingLocation(timestamp, "blockSamples"));
```
sonalgoyal, line 65: No - we need to pick the top 3 hashes by count and print their samples. this logic is wrong

sonalgoyal (changes requested):
see comments. also look at pmd warnings and fix those please

common/core/src/main/java/zingg/common/core/block/Blocker.java
sania-16: so we first sort the hashes in descending order, the pick the top 3 and print them?
we do not show 10% of those top 3?

common/core/src/main/java/zingg/common/core/block/Blocker.java
```diff
@@ -0,0 +1,90 @@
...
+import zingg.common.client.pipe.FilePipe;
+import zingg.common.client.pipe.Pipe;
+import zingg.common.client.util.ColName;
+import zingg.common.core.executor.ZinggBase;
+
+public class Blocker<S,D,R,C,T> extends ZinggBase<S,D,R,C,T>{
+
+	private static final long serialVersionUID = 1L;
+	protected static String name = "zingg.Blocker";
+	public static final Log LOG = LogFactory.getLog(Blocker.class);
```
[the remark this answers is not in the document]
sania-16: @sonalgoyal
what needs to be changed here - have used the same code as in Matcher?

common/core/src/main/java/zingg/common/core/block/Blocker.java
sonalgoyal: we have to print two things
1. hashes by count
2. for the top 3 hashes by count, we need to sample 10% of the data nad write that

So say
hash rec
1 r1
1 r2
1 r3
2 r4
2 r5
2 r6
2 r7
2 r8
2 r9
2 r10
2 r11


So we will have
hash count
1 3
2 7

Now if we were doing top 1 , we will sample from hash 2 10% records and write them. We are doing top 3 so build it accordingly

sania-16: how do we iterate through the ZFrame?

sania-16:
@sonalgoyal 
have tested on febrl and febl120k 
need to structure junits

common/core/src/main/java/zingg/common/core/executor/Matcher.java
```diff
@@ -46,6 +47,9 @@ public ZFrame<D, R, C> getFieldDefColumnsDS(ZFrame<D, R, C> testDataOriginal) {
 
 
 	public ZFrame<D,R,C>  getBlocked( ZFrame<D,R,C>  testData) throws Exception, ZinggClientException{
+		//for blocking
+		Blocker<S,D,R,C,T> blocker = new Blocker<S,D,R,C,T>();
+        blocker.getBlocked(testData);
```
sania-16, line 52: in this part - do we need the logic of getBlocked again or we just call the getBlocked from Blocker and that's it? @sonalgoyal

sania-16:
for junit - test counts are same in test output and given set input which has z_hash also based on test csv
            
            //read output pipe - counts
            //input data has z_hash also as a column
            //find the counts in input data and arrange in desc order
            //asserts true on input and output
            
does this look fine @sonalgoyal ?

common/core/src/test/java/zingg/common/core/executor/BlockerValidator.java
```diff
@@ -0,0 +1,29 @@
...
+
+	@Override
+	public void validateResults() throws ZinggClientException {
+			//test counts are same in test output and given set input which has only z_hash also based on example test
+            
+            //read output pipe - counts
+            //input data has z_hash also as a column
+            //find the counts in input data and arrange in desc order
+            //asserts true on input and output
+    }
```
sania-16, line 28: @sonalgoyal 
please review the logic for junit - will make changes accordingly

sania-16:
have added docs here

common/core/src/main/java/zingg/common/core/executor/BlockingTreeDebugger.java
```diff
@@ -0,0 +1,35 @@
...
+import zingg.common.client.util.ColName;
+import zingg.common.core.block.Blocker;
+import zingg.common.core.block.Canopy;
+import zingg.common.core.block.Tree;
+
+public abstract class BlockingTreeDebugger<S,D,R,C,T> extends ZinggBase<S,D,R,C,T> {
+
+    private static final long serialVersionUID = 1L;
+	protected static String name = "zingg.BlockingTreeDebugger";
+	public static final Log LOG = LogFactory.getLog(BlockingTreeDebugger.class);
```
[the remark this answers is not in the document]
sonalgoyal: rename

sonalgoyal (changes requested):

common/core/src/main/java/zingg/common/core/block/Blocker.java
```diff
@@ -12,13 +11,11 @@
...
 import zingg.common.core.executor.ZinggBase;
-import zingg.common.core.sink.TableOutput;
-import zingg.common.core.util.Analytics;
-import zingg.common.core.util.BlockingTreeUtil;
-import zingg.common.core.util.Metric;
 
 public class Blocker<S,D,R,C,T> extends ZinggBase<S,D,R,C,T>{
 
+	private static final long serialVersionUID = 1L;
+	protected static String name = "zingg.Blocker";
```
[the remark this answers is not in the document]
sania-16: BlockVerifier - this works?

common/core/src/main/java/zingg/common/core/block/Blocker.java
```diff
@@ -0,0 +1,121 @@
...
+
+    public ZFrame<D,R,C> getBlocked(ZFrame<D,R,C> testData) throws Exception, ZinggClientException{
+		LOG.debug("Blocking model file location is " + args.getBlockFile());
+		Tree<Canopy<R>> tree = getBlockingTreeUtil().readBlockingTree(args);
+		ZFrame<D,R,C> blocked = getBlockingTreeUtil().getBlockHashes(testData, tree);
+		ZFrame<D,R,C> blocked1 = blocked.repartition(args.getNumPartitions(), blocked.col(ColName.HASH_COL)).cache();
+		return blocked1;
+	}
+
+    public Pipe<D,R,C> getPipeForVerifyBlockingLocation(long timestamp, String type){
```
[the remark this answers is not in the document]
sania-16: this will be in executor itself?

common/core/src/main/java/zingg/common/core/executor/BlockingTreeDebugger.java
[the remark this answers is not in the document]
sania-16: this will be renamed to Blocker?

docs/verifyBlocking.md
```diff
@@ -0,0 +1,14 @@
...
+
+# Verification of Blocking Data
+
+Sometimes Zingg jobs are slow or fail due to a poorly learnt blocking tree. This can happen due to a variety of reasons. It can happen when a user adds significantly larger training samples compared to the labelling learnt by Zingg or due to a natural bias in the data with lots of null columns used in matching. If we have an understanding of how blocking is working before deciding to run a match or link job, we get a better idea whether we need to add more training data for getting better results.
+
+### The verifyBlocking phase is run as follows:
+
+`./scripts/zingg.sh --phase verifyBlocking --conf <path to conf> <optional --zinggDir <location of model>>`
+
+The output contains two directories - zinggDir/modelId/blocks/timestamp/counts and zinggDir/modelId/blocks/timestamp/blockSamples. We can see the counts per block and the top 10% records associated with the top 3 blocks by counts in the directories respectively.
```
sonalgoyal, line 14: define Enterprise Snowflake how it will work as well

docs/verifyBlocking.md
```diff
@@ -0,0 +1,14 @@
+---
+description: >-
+  Understanding how blocking is working before running match or link
+---
+
+# Verification of Blocking Data
+
+Sometimes Zingg jobs are slow or fail due to a poorly learnt blocking tree. This can happen due to a variety of reasons. It can happen when a user adds significantly larger training samples compared to the labelling learnt by Zingg or due to a natural bias in the data with lots of null columns used in matching. If we have an understanding of how blocking is working before deciding to run a match or link job, we get a better idea whether we need to add more training data for getting better results.
```
sonalgoyal, line 8: It can happen when 
- a user adds significantly larger training samples compared to the labelling learnt by Zingg. The manually added training samples may have the same type of columns and blocking rules learnt are not generic enough. For example providing California state only training data when the matching is using the State column and data has multiple states.
- When there is a natural bias in the data with lots of null columns used in matching. 
- When sufficient labeling has not been done.
- Having lots of non differentiating columns

sonalgoyal: Also define later how to use the information

common/core/src/test/java/zingg/common/core/executor/BlockerValidator.java
sonalgoyal: yes looks ok to me. you can unit test the methods which return the counts etc

sonalgoyal (changes requested):

sania-16:
@sonalgoyal 
just wanted to confirm the individual junits once
for VerifyBlockingPipes - check whether pipe is being formed
for Blocker - check the blocks being returned
for VerifyBlocking - check the counts being returned

common/core/src/main/java/zingg/common/core/block/Blocker.java
```diff
@@ -0,0 +1,28 @@
...
+public class Blocker<S,D,R,C,T> {
+
+    private static final long serialVersionUID = 1L;
+	private static final Log LOG = LogFactory.getLog(Blocker.class);
+
+    public Blocker(){
+    }
+
+    public ZFrame<D,R,C> getBlocked(ZFrame<D,R,C> testData, IArguments args, BlockingTreeUtil<S,D,R,C,T> inputTree ) throws Exception, ZinggClientException{
+		LOG.warn("Blocking model file location is " + args.getBlockFile());
```
sonalgoyal, line 21: remove "file" from the log since it is equally applicable to snowflake tables.

common/core/src/main/java/zingg/common/core/executor/verifyBlocking/VerifyBlocking.java
```diff
@@ -0,0 +1,99 @@
...
+
+		for(R row: topRec) {
+			int hash = (int) blockTopRec.get(row, ColName.HASH_COL);
+			long count = (long) blockTopRec.get(row, ColName.HASH_COL + "_count");
+			int sampleSize = Math.max(1, (int) Math.ceil(count * 0.1));
+			ZFrame<D,R,C> matchingRecords = null;
+
+			for(R rows: dataRec)
+			{
+				int recHash = (int)blocked.get(rows,ColName.HASH_COL);
```
sonalgoyal, line 70: hash is a long

common/core/src/main/java/zingg/common/core/executor/verifyBlocking/VerifyBlocking.java
```diff
@@ -0,0 +1,99 @@
...
+
+	public void getBlockSamples(ZFrame<D, R, C> blocked, ZFrame<D, R, C> blockTopRec) throws ZinggClientException {
+		List<R> topRec = blockTopRec.collectAsList();
+		List<R> dataRec = blocked.collectAsList();
+
+
+		for(R row: topRec) {
+			int hash = (int) blockTopRec.get(row, ColName.HASH_COL);
+			long count = (long) blockTopRec.get(row, ColName.HASH_COL + "_count");
+			int sampleSize = Math.max(1, (int) Math.ceil(count * 0.1));
```
sonalgoyal, line 65: you can simply get the 0.1 fromt he original ungrouped blockedDf

common/core/src/main/java/zingg/common/core/executor/verifyBlocking/VerifyBlocking.java
```diff
@@ -0,0 +1,99 @@
...
+			}
+			e.printStackTrace();
+			throw new ZinggClientException(e.getMessage());
+		}
+    
+    }
+
+	public void getBlockSamples(ZFrame<D, R, C> blocked, ZFrame<D, R, C> blockTopRec) throws ZinggClientException {
+		List<R> topRec = blockTopRec.collectAsList();
+		List<R> dataRec = blocked.collectAsList();
```
sonalgoyal, line 59: you can not do this. blocked will be a huge df with millions of records. collect on blockTopRec, get the 3 hashes, then filter on blocked

common/core/src/main/java/zingg/common/core/executor/verifyBlocking/VerifyBlocking.java
```diff
@@ -0,0 +1,99 @@
...
+import zingg.common.core.block.InputDataGetter;
+import zingg.common.core.executor.ZinggBase;
+
+public abstract class VerifyBlocking<S,D,R,C,T> extends ZinggBase<S,D,R,C,T>{
+
+	private static final long serialVersionUID = 1L;
+	protected static String name = "zingg.VerifyBlocking";
+	public static final Log LOG = LogFactory.getLog(VerifyBlocking.class);
+    public long timestamp = System.currentTimeMillis();   
+    VerifyBlockingPipes<S,D,R,C> getPipe = new VerifyBlockingPipes<S,D,R,C>();
```
sonalgoyal, line 23: rename to verifyBlockingPipes

common/core/src/main/java/zingg/common/core/executor/verifyBlocking/VerifyBlocking.java
```diff
@@ -0,0 +1,99 @@
...
+					matchingRecords = blocked.select(blocked.getCols()).withColumn(ColName.HASH_COL,hash).limit(sampleSize);
+				}
+			}
+			
+			getPipeUtil().write(matchingRecords, getPipe.getPipeForVerifyBlockingLocation(args, getPipeUtil(),timestamp, "blockSamples/" + hash));
+		}
+		
+	}
+
+	public ZFrame<D, R, C> getFieldDefColumnsDS(ZFrame<D, R, C> testDataOriginal) {
```
sonalgoyal, line 81: why is this needed here?

common/core/src/main/java/zingg/common/core/executor/verifyBlocking/VerifyBlockingPipes.java
```diff
@@ -0,0 +1,27 @@
+package zingg.common.core.executor.verifyblocking;
+
+import zingg.common.client.IArguments;
+import zingg.common.client.pipe.FilePipe;
+import zingg.common.client.pipe.Pipe;
+import zingg.common.client.util.PipeUtilBase;
+
+public class VerifyBlockingPipes<S,D,R,C>  {
```
sonalgoyal, line 8: this needs to be an interface. the actual implementation that you have defined is spark specific and would go there

sonalgoyal (commented):
please change

sania-16:
make number of records and percentage of records as class property

sania-16:
@sonalgoyal 
not sure how to get matching records without iteration - tried multiple methods for the same and was not able to do so
have made the remaining changes - please review

common/client/src/main/java/zingg/common/client/ArgumentsUtil.java
```diff
@@ -104,7 +104,7 @@ public void writeArgumentsToJSON(String filePath, IArguments args)
 	}
 
 	public void checkValid(IArguments args, String phase) throws ZinggClientException {
-		if (phase.equals("train") || phase.equals("match") || phase.equals("trainMatch") || phase.equals("link")) {
+		if (phase.equals("train") || phase.equals("verifyBlocking") || phase.equals("match") || phase.equals("trainMatch") || phase.equals("link") ) {
```
Nitish1814, line 107: can you please create some Enum class from where these constants can be taken up?, enum class name can be decided
```
public enum RequiredPhase {
    TRAIN("train"),
    MATCH("match");

    private final String phase;

    RequiredPhase(String phase) {
        this.phase = phase;
    }

    public boolean checkIfPresent(String phase) {
        for (RequiredPhase requiredPhase : RequiredPhase.values()) {
            if (requiredPhase.phase.equals(phase)) {
                return true;
            }
        }
        return false;
    }

    public boolean checkIfPresent(RequiredPhase phase) {
        for (RequiredPhase requiredPhase : RequiredPhase.values()) {
            if (requiredPhase.equals(phase)) {
                return true;
            }
        }
        return false;
    }
}
```

sonalgoyal: Let us not do this. The methods checkValid etc are going to be cleaned up. It is an open issue.

Nitish1814: ok, then it should be fine if we are planning for cleanup later

PR #932, opened 2024-10-29
added singleton SparkSessionProvider

Nitish1814:


spark/core/src/test/java/zingg/spark/core/TestImageType.java
```diff
@@ -91,7 +92,7 @@ public void testUDFArray() {
 			df.printSchema();
 			// register ArrayDoubleSimilarityFunction as a UDF
 			TestUDFDoubleArr testUDFDoubleArr = new TestUDFDoubleArr();
-			SparkFnRegistrar.registerUDF2(spark, "testUDFDoubleArr", testUDFDoubleArr, DataTypes.DoubleType);
+			SparkFnRegistrar.registerUDF2(TestSparkBase.spark, "testUDFDoubleArr", testUDFDoubleArr, DataTypes.DoubleType);
```
sonalgoyal, line 95: why not inject spark instance variqable and not make changes throughout? The whole idea is the class code should not know or care about how spark session was created.

spark/core/src/test/java/zingg/spark/core/documenter/TestDocumenterBase.java
```diff
@@ -14,21 +14,34 @@
...
+import zingg.common.client.ArgumentsUtil;
+import zingg.common.client.IArguments;
+import zingg.common.core.documenter.DocumenterBase;
+import zingg.common.core.documenter.TemplateFields;
+import zingg.spark.core.TestSparkBase;
 
 
-public class TestDocumenterBase extends ZinggSparkTester {
+@ExtendWith(TestSparkBase.class)
+public class TestDocumenterBase {
```
sonalgoyal, line 35: we should have a ganeric test and a spark specific test for these classes

sonalgoyal (changes requested):

spark/core/src/test/java/zingg/spark/core/TestImageType.java
Nitish1814: updated

spark/core/src/test/java/zingg/spark/core/documenter/TestDocumenterBase.java
Nitish1814: issue created

PR #933, opened 2024-10-29
exception thrown if data type not supported

Nitish1814:


sonalgoyal:
@LOGANBLUE1 please check

PR #938, opened 2024-11-06
Laod test suite

Nitish1814:


perf_test_input.py
```diff
@@ -0,0 +1,19 @@
+#phases to test
+FIND_TRAINING_DATA = "findTrainingData"
+LABEL = "label"
+MATCH = "match"
+INCREMENTAL = "runIncrement"
+
+#load file config to test on
+LOAD_65 = "/home/administrator/zingg/zinggEnterprise/spark/examples/febrl/config.json"
```
sonalgoyal, line 8: why do we need to give locations from root? also referncing enterprise stuff here? /home/adminsitroator should not be there.

perf_test_input.py
```diff
@@ -0,0 +1,19 @@
...
+LOAD_120k = "/home/administrator/zingg/zinggEnterprise/spark/examples/febrl120k/config120k.json"
+
+
+#bash script location
+ZINGG = "/home/administrator/zingg/zinggEnterprise/spark/scripts/zingg.sh"
+
+
+#add all the load to test
+load_configs = {LOAD_65 : "LOAD_65"}
```
sonalgoyal, line 17: please add all the relevant tests we discussed

.github/workflows/load-test.yml
```diff
@@ -0,0 +1,25 @@
+name: load-test
+
+on:
+  schedule:
+    - cron: "*/15 * * * *"
```
sonalgoyal, line 5: has to be once every 3 days

loadTest.py
```diff
@@ -0,0 +1,106 @@
...
+    print("Running phase - " + phase)
+    return subprocess.run(ZINGG + " %s %s %s %s" % ("--phase", phase, "--conf", conf), stderr=PIPE, stdout=PIPE, shell=True)
+
+def perf_test_small(phase):
+    return "small_test_running"
+
+
+def write_on_start():
+    f = open("loadTestReport_" + str(start_time), "w+")
+    f.write("******************************** perf test report, " + str(date.today()) + " ********************************\n\n");
```
sonalgoyal, line 30: try to use format for the test report - it will be much cleaner and easier to maintain. will also prevent spacing and tab issues between header and value

.github/workflows/load-test.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  load-test:
+    name: load-test-spark
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout repo content
```
sonalgoyal, line 12: how does zingg get compiled, env set etc?

sonalgoyal (commented):
see comments

PR #952, opened 2024-11-12
Laod test suite

Nitish1814:
changes
1- added time to report
2- report will be created under ./perfTestReport folder
3- every report will overwrite previous report


.github/workflows/load-test.yml
```diff
@@ -1,6 +1,7 @@
 name: load-test
 
 on:
+  push:
```
sonalgoyal, line 4: why on push?

common/core/src/main/java/zingg/common/core/block/Canopy.java
```diff
@@ -16,6 +16,7 @@
 
 
 public class Canopy<R> implements Serializable {
+	private static final long serialVersionUID = -229533781044789499L;
```
sonalgoyal, line 19: why are we adding it here? and this value in any case is not making sense.

perf_test_input.py
```diff
@@ -7,13 +7,14 @@
 #load file config to test on
 config_65 = "examples/febrl120k/config.json"
```
sonalgoyal, line 8: rename to febrl, febrl120k, ncVoter5m

sonalgoyal (changes requested):
report name has date which means it is not getting overwritten?

see other comments.

perf_test_input.py
Nitish1814: done

PR #954, opened 2024-11-13
Refactoring code for junits

sania-16:


sania-16:
need to include teardown
need to update accuracy tests

PR #971, opened 2024-11-26
Verify blocking Refactoring

sania-16:


common/core/src/main/java/zingg/common/core/executor/blockingverifier/VerifyBlocking.java
```diff
@@ -89,6 +89,16 @@ public void setVerifyBlockingPipeUtil(IVerifyBlockingPipes<S,D,R,C> verifyBlocki
 		this.verifyBlockingPipeUtil = verifyBlockingPipeUtil;
 	}
 
+	public ZFrame<D,R,C> getTestDataOriginal() throws ZinggClientException{
```
sonalgoyal, line 92: why is this testDataOriginal and not testData ?

common/core/src/main/java/zingg/common/core/executor/blockingverifier/VerifyBlocking.java
```diff
@@ -89,6 +89,16 @@ public void setVerifyBlockingPipeUtil(IVerifyBlockingPipes<S,D,R,C> verifyBlocki
...
+		ZFrame<D,R,C> testData = new InputDataGetter<S,D,R,C>(getPipeUtil()).getTestData(args);
+		return testData;
+	}
+
+	public ZFrame<D,R,C> getBlockedData(ZFrame<D,R,C> testDataOriginal) throws ZinggClientException, Exception{
+		ZFrame<D,R,C> blockedData = new Blocker<S,D,R,C,T>(getBlockingTreeUtil()).getBlocked(testDataOriginal,args, getModelHelper());
```
sonalgoyal, line 98: you should be calling getBlocker().getBlocked here and getBlocker gets overridden in child classes

sonalgoyal (changes requested):
minor fixes

Issue #918, opened 2024-10-18
improve look and feel of generated docs


sonalgoyal:
Column width should be auto deduced. Sometimes the valeus are overlapping
![image](https://github.com/user-attachments/assets/36b36d4e-3362-46f6-abc7-b14dca92973b)

Color scheming can be better

We can show pos and neg separately, right now they are appearing as per the order of the labeling rounds

PR #979, opened 2024-12-05
Generate Docs - changes

sania-16:
issue #918


common/core/src/main/resources/model.ftlh
```diff
@@ -19,17 +19,20 @@
 	  </tbody>
   </table>
  <p>
-  <table class="table table-borderless">
+  <table class="table table-hover">
+  <table class="table table-striped">
```
sonalgoyal, line 23: you dont need two tables?

sonalgoyal (commented):
why do oyu need two tables?

common/core/src/main/resources/model.ftlh
sania-16: missed it -will remove

PR #1001, opened 2025-01-14
VerifyBlocking PR

sania-16:


common/client/src/main/java/zingg/common/client/util/PathUtil.java
```diff
@@ -0,0 +1,20 @@
+package zingg.common.client.util;
+
+public class PathUtil {
+
+    public static String getSparkLocationFromPath(String name){
```
sonalgoyal, line 5: since these are spark and snow specific, it is better to move these methods to SnowPipe and SparkPipe. call them massageLocation/massageTableName. define method setLocation in spark and call the massageLocation in it. Use setLocation whever we are calling SparkPipe.setLocation. Similarly for Snow.

sonalgoyal (changes requested):
check comment

PR #1013, opened 2025-01-21
added IZinggSuper interface

Nitish1814:


common/client/src/main/java/zingg/common/client/IZinggSuper.java
```diff
@@ -0,0 +1,19 @@
+package zingg.common.client;
+
+//TODO need to revisit this interface
+public interface IZinggSuper<S,D,R,C> {
```
sonalgoyal, line 4: can you please rename this to IZinggModelInfo or some such thing?

PR #1027, opened 2025-02-06
Case normalize

Nitish1814:


common/core/src/main/java/zingg/common/core/preprocess/caseNormalize/CaseNormalizer.java
```diff
@@ -0,0 +1,88 @@
+package zingg.common.core.preprocess.caseNormalize;
```
sonalgoyal, line 1: java packages should have all lowercase. we messed up at infraForTest but lets follow the convention

common/core/src/main/java/zingg/common/core/preprocess/caseNormalize/CaseNormalizer.java
```diff
@@ -0,0 +1,88 @@
...
+import zingg.common.client.MatchTypes;
+import zingg.common.client.ZFrame;
+import zingg.common.client.ZinggClientException;
+import zingg.common.core.context.IContext;
+import zingg.common.core.preprocess.IMultiFieldPreprocessor;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class CaseNormalizer<S,D,R,C,T> implements IMultiFieldPreprocessor<S,D,R,C,T> {
```
sonalgoyal, line 15: better to break it down into one abstract class AMultiFieldPreproc and then extend it. So other multi preprocs would not need to define any of the getter setters

common/core/src/main/java/zingg/common/core/preprocess/caseNormalize/CaseNormalizer.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public IContext<S, D, R, C, T> getContext() {
+        return this.context;
+    }
+
+    @Override
+    public boolean isApplicable() {
+        return true;
```
sonalgoyal, line 51: should return true only when you have string fields

common/core/src/main/java/zingg/common/core/preprocess/caseNormalize/CaseNormalizer.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public List<? extends FieldDefinition> getFieldDefinitions() {
+        return this.fieldDefinitions;
+    }
+
+    private List<String> getRelevantFields() {
+        List<String> stringFields = new ArrayList<>();
+        for (FieldDefinition fieldDefinition : fieldDefinitions) {
+            if (fieldDefinition.dataType != null && fieldDefinition.matchType != null &&
+                    fieldDefinition.dataType.equalsIgnoreCase(STRING_TYPE) && !fieldDefinition.matchType.contains(MatchTypes.DONT_USE)) {
```
sonalgoyal, line 80: lets discuss this. so far we have always abstracted dataType

sonalgoyal (changes requested):
looks good overall. minor comments.

sonalgoyal (changes requested):
minor changes

PR #1032, opened 2025-02-08
Generic Tests PR

sania-16:


common/core/src/test/java/zingg/common/core/recommender/data/TestStopWordRecommenderData.java
```diff
@@ -0,0 +1,25 @@
+package zingg.common.core.recommender.data;
+
+import java.util.ArrayList;
+import java.util.List;
+import zingg.common.core.recommender.model.WordByCount;
+
+public class TestStopWordRecommenderData {
```
sonalgoyal, line 7: remove test from class name

PR #1062, opened 2025-02-25
Performance test issue#1052

Arjun-Zingg:
Generates test report in json format.
Fails the test before saving the results if the performance drops by 10% or more.

### DIFF
--- a/perf_test/perfTestRunner.py
+++ b/perf_test/perfTestRunner.py
@@ -15,7 +15,7 @@ current_time = now.strftime("%H:%M:%S")
 reportFile = "./perf_test/perf_test_report/loadTestReport.json"
 
 propertyFile = "./config/zingg.conf"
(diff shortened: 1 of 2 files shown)


perf_test/perfTestRunner.py
```diff
@@ -1,118 +1,139 @@
...
+    compare_results(prev_results, phase_time)
 
-        #write success data to file
-        if phase_time:
-            write_success_stats(phase_time, load)
-        #write failure data to file
-        if phase_error:
-            write_failure_stats(phase_error)
+    # Save only if no performance degradation was detected
+    test_data["results"] = {}
```
sonalgoyal, line 110: you have to save irrespective of change, but if there a degradation of more than 5%, please fail

sonalgoyal (changes requested):

PR #1136, opened 2025-04-09
Telemetry changes

sania-16:


common/core/src/main/java/zingg/common/core/util/Analytics.java
sonalgoyal, line 1: It has to stay at warm level

log4j2.properties
sonalgoyal, line 1: Make it off

perf_test/perf_test_report/testReport_febrl120K.json
sonalgoyal, line 1: These files should not be checked in and should not be part of any pr

sonalgoyal (changes requested):

perf_test/perf_test_report/testReport_febrl120K.json
sania-16: added to .gitignore

PR #1149, opened 2025-04-28
pass arguments to getBlocks

Nitish1814:


spark/core/src/main/java/zingg/spark/core/util/SparkBlockingTreeUtil.java
```diff
@@ -86,7 +87,7 @@ public Tree<Canopy<Row>> readBlockingTree(Arguments args) throws Exception, Zing
 public Block<Dataset<Row>, Row, Column, DataType> getBlock(ZFrame<Dataset<Row>, Row, Column> sample,
                                                            ZFrame<Dataset<Row>, Row, Column> positives,
                                                            ListMap<DataType, HashFunction<Dataset<Row>, Row, Column, DataType>> hashFunctions,
-                                                           long blockSize) {
+                                                           long blockSize, IArguments arguments) {
         // TODO Auto-generated method stub
         return new SparkBlock(sample, positives, hashFunctions, blockSize);
```
sonalgoyal, line 92: i would suggest changing this constructor as well.

sonalgoyal (changes requested):

spark/core/src/main/java/zingg/spark/core/util/SparkBlockingTreeUtil.java
Nitish1814: updated

PR #1188, opened 2025-08-06
added default argument loader

Nitish1814:


common/client/src/main/java/zingg/common/client/arguments/loader/ArgumentsLoader.java
```diff
@@ -15,5 +17,9 @@ public ArgumentsLoader(Class<A> argsClass) {
                 .configure(JsonParser.Feature.ALLOW_UNQUOTED_CONTROL_CHARS, true);
     }
 
-    public abstract A load(String path) throws ZinggClientException;
+    public abstract A load(String path) throws ZinggClientException, InstantiationException, IllegalAccessException, NoSuchMethodException, InvocationTargetException;
+
+    public A load() throws NoSuchMethodException, InvocationTargetException, InstantiationException, IllegalAccessException {
```
sonalgoyal: this is not what load is supposed to do - here we are creating a blank Args class which is not really loading

sonalgoyal: should not be part of oss at all

sonalgoyal: create an interface and code to it for the service

common/client/src/main/java/zingg/common/client/arguments/loader/DefaultArgumentLoader.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.lang.reflect.InvocationTargetException;
+
+public class DefaultArgumentLoader<A extends IZArgs> extends ArgumentsLoader<A> {
+    public DefaultArgumentLoader(Class<A> argsClass) {
+        super(argsClass);
+    }
+
+    @Override
+    public A load(String path) throws ZinggClientException, InstantiationException, IllegalAccessException, NoSuchMethodException, InvocationTargetException {
```
sonalgoyal: wrong practise of passing arg and not using it in the method

sonalgoyal (changes requested):

common/client/src/main/java/zingg/common/client/arguments/loader/ArgumentsLoader.java
Nitish1814: upadated, this interface earlier had overloaded load() method for creating default argument object without any properties set. But now we don't need it

Nitish1814: upadated, this interface earlier had overloaded load() method for creating default argument object without any properties set. But now we don't need it

common/client/src/main/java/zingg/common/client/arguments/loader/DefaultArgumentLoader.java
Nitish1814: It was made as default method of abstract class(so, it was not breaking Liskov's substitution) but anyways removed it.

PR #1190, opened 2025-08-11
0.5.1 webhooknotify

padam-prakash:


common/client/src/main/java/zingg/common/client/Client.java
```diff
@@ -217,41 +218,40 @@ else if (options.get(ClientOptions.CONF).value.endsWith("env")) {
...
+				}
 			}
 			catch(ZinggClientException e) {
 				if (options != null && options.get(ClientOptions.EMAIL) != null) {
 					Email.email(options.get(ClientOptions.EMAIL).value, new EmailBody("Error running Zingg job",
 						"Zingg Error ",
 						e.getMessage()));
 				}
+				if (!success) {
+					System.exit(1);
```
sonalgoyal: fail event has to be fired here as well. thats why we need a fail method that takes success and does

```
 if (!success) {
					EventsListener.getInstance().fireEvent(new ZinggFailEvent());
					System.exit(1);
```

sonalgoyal (changes requested):

PR #1213, opened 2025-10-27
Issue 1084 document update for inc

padam-prakash:
improve incremental flow documentation, clarify outputTmp usage, and restructure explanation for readability

### DIFF
--- a/docs/runIncremental.md
+++ b/docs/runIncremental.md
@@ -11,7 +11,7 @@ description: >-
 
 [Zingg Enterprise Feature](#user-content-fn-1)[^1]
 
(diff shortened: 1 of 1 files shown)


docs/runIncremental.md
```diff
@@ -104,6 +116,8 @@ incrArgs = IncrementalArguments()
 incrArgs.setParentArgs(args)  
 incrPipe = ECsvPipe("testFebrlIncr", "examples/febrl/test-incr.csv", schema)  
 incrArgs.setIncrementalData(incrPipe)  
+outputTempPipe = ECsvPipe("outputTemp", "/tmp/zinggOutput_febrl_tmp")
```
sonalgoyal: has to have headers?

padam-prakash: Not necessarily. Both True and False values for outputTmp header work. If the header is set to True for outputTmp, the final output will include the header.

sonalgoyal: i think we realized that we need it to have headerselse the output goes bad?

docs/runIncremental.md
```diff
@@ -11,14 +11,17 @@ description: >-
...
+Rerunning matching on entire datasets is wasteful, and we lose the lineage of matched records against a persistent identifier. Using the incremental flow feature in Zingg Enterprise, incremental loads can be run to match existing pre-resolved entities. The new and updated records are matched to existing clusters, and new persistent ZINGG_IDs are generated for records that do not find a match.
```
sonalgoyal: links are removed?

docs/runIncremental.md
padam-prakash: added headers

docs/runIncremental.md
padam-prakash: added the updated links

PR #1260, opened 2026-01-18
Nbl/docs macos setup

nabeel054002:
This contains the setting up steps for development environment for Mac OS users.
Our current setup contains setting up instructions for linux environments.

### DIFF
--- a/docs/macOSSetup.md
+++ b/docs/macOSSetup.md
@@ -120,8 +120,39 @@ mvn clean compile package -Dspark=sparkVer -Dmaven.test.skip=true
 
 ---
 
+### **Step 8: Run Zingg To Find Training Data**
+
+* Run this script in the terminal opened in Zingg clones directory:
+```bash
+./scripts/zingg.sh --phase findTrainingData --conf examples/febrl/config.json
+```
+
+**If everything is right, it should show Zingg banner.**
+
 ---
 
-### **Next Steps**
+### **Step 9: Run Zingg To Label Data**
 
-After completing the macOS-specific setup above, refer to the [main setup guide](./settingUpZingg.md) for Steps 8-11 to run Zingg (findTrainingData, label, train, and match phases).
+* Run this script in the terminal opened in Zingg clones directory:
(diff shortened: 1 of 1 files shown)


docs/settingUpZingg.md
```diff
@@ -1,8 +1,18 @@
...
 
-The following steps will help you set up the Zingg Development Environment. While the steps remain the same across different OS, we have provided detailed instructions for Ubuntu OS. \
-\
-The below steps have been created using Ubuntu 22.04.2 LTS
+The following steps will help you set up the Zingg Development Environment. While the core steps remain the same across different OS, we have provided detailed instructions for **macOS** and **Ubuntu/WSL2**.
+
+### **Step 0: Initial OS Setup**
+
+#### **For macOS**
+Homebrew is required to install system dependencies.
```
inshane13: instead of mentioning homebrew is required, you can say like homebrew is preferable, it makes installation easy

nabeel054002: it is used to install other dependancies, should i only change this particular sentence or provide any alternate method for installation as well?

inshane13: it should be flexible for users and guide them though easy way out

nabeel054002: have updated the pre brew installation text, pls do take a look thanks

docs/macOSSetup.md
```diff
@@ -32,6 +38,23 @@ brew install git
 
 ---
 
+### **Step 3: Install Apache Spark**
+
+* Download Apache Spark from the [Apache Spark Official Website](https://spark.apache.org/downloads.html).
+* For example for 3.5.0:
```
inshane13, line 44: zingg runs on spark 3.5.5 as well, if you like to install the latest, its better

nabeel054002: Got it, although if its optional, can I do away by writing an additional note on the above as the installation processes for the two are the same

inshane13 (approved):

PR #1265, opened 2026-01-29
Fix exception wrapping to retain original message

LOGANBLUE1:
In this PR we fixed the usage of ZinggClientException, preserved error messages.
Fixed under issue - [link](https://github.com/zinggAI/zinggEnterprise/issues/1151)

### DIFF
--- a/assembly/src/assembly/dist.xml
+++ b/assembly/src/assembly/dist.xml
@@ -44,6 +44,18 @@
       <includes>
         <include>**/*</include>
       </includes>
+      <excludes>
+        <exclude>venv/**</exclude>
+        <exclude>.venv/**</exclude>
+        <exclude>**/__pycache__/**</exclude>
+        <exclude>**/*.pyc</exclude>
+        <exclude>**/*.pyo</exclude>
+        <exclude>build/**</exclude>
+        <exclude>dist/**</exclude>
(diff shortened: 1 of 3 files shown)


common/core/src/main/java/zingg/common/core/util/ModelUtil.java
```diff
@@ -45,11 +45,11 @@ public void loadFeatures(IArguments args) throws ZinggClientException {
 			LOG.info("Finished reading internal configurations and functions");
 			}
 		}
-		catch(Throwable t) {
+		catch(Exception e) {
 			LOG.warn("Unable to initialize internal configurations and functions");
-            t.printStackTrace();
-			if (LOG.isDebugEnabled()) t.printStackTrace();
```
sonalgoyal, line 51: you should print stack trace only in denug scnearios. stacktrace is getting printed otherwise in client

common/core/src/main/java/zingg/common/core/util/ModelUtil.java
```diff
@@ -45,11 +45,11 @@ public void loadFeatures(IArguments args) throws ZinggClientException {
...
-			throw new ZinggClientException("Unable to initialize internal configurations and functions");
+            e.printStackTrace();
```
sonalgoyal: ?

Issue #1175, opened 2025-07-02
release tar is too big


sonalgoyal:
we should look at separating the model and examples into a different repo maybe

sonalgoyal:
We can simplify this by including only the febrl 65 and 120k examples in the release

PR #1266, opened 2026-01-30
tar compression

inshane13:
I have removed all examples and models from the release tar expcept febrl 120k and its model.
here are the results. closes: https://github.com/zinggAI/zingg/issues/1175
before:
Screenshot 2026-01-30 at 17 28 35
after:
Screenshot 2026-01-30 at 17 28 46


sonalgoyal:
it is still v big. can you please tell me whats so huge here?

also, have you tested the changed release by unzipping and installing?

PR #1277, opened 2026-02-18
removed try-catch from the junits

LOGANBLUE1:


spark/core/src/test/java/zingg/spark/core/TestDocumenter.java
padam-prakash, line 1: add test documenter test cases later

padam-prakash (approved):
Approved

PR #1333, opened 2026-06-30
blockingTree performance improvements - Added benchmarks for different optimisations

LOGANBLUE1:
Command to run benchmarks: mvn integration-test -Pbenchmark -pl spark/core -am -DskipTests
Output file path: zingg/spark/core/target/benchmark-results.json

Optimizations that are included -

1. Reuse a probe Canopy instead of allocating per candidate
getNodeFromCurrent runs for every (field × function) candidate — with 10 fields × N functions that's potentially 50+ new Canopy + copyTo calls per getBestNode invocation, all immediately discarded. Only the winner needs a real object.
Fix: allocate one probe Canopy at the start of getBestNode, set function and context on it per iteration (both are just field assignments from copyTo), then countEliminationsWithValues and estimateCanopies run on it. Only when a new best is found, materialize a real Canopy with getNodeFromCurrent. This cuts N allocations per call down to 1.

2. Pass precomputed values into buildDupeRemaining
buildDupeRemaining calls function.apply(r, context.fieldName) per row — full row field extraction — even though getBestNode already has preVals1/preVals2 for that exact field. The problem is those arrays are local to the field loop and buildDupeRemaining is called after the loop ends, when the winning field's preVals are out of scope.
Fix: when a new best is found inside the loop, snapshot bestPreVals1 and bestPreVals2 alongside it. Pass them to buildDupeRemaining so it can call function.applyToValue(bestPreVals1[i]) instead of function.apply(r, fieldName) — same savings as countEliminationsWithValues already gets.

3. Precompute field values + early exit in estimateElimCount
estimateElimCount calls function.apply(r, context.fieldName) and function.apply(r, COL_PREFIX + context.fieldName) for every row in dupeN, for every candidate — with 10 fields × N functions × dupeN rows, the same field value is extracted from the Row repeatedly across functions in the same field. Additionally, it builds dupeRemaining for every evaluated candidate, including all the losers that are immediately discarded.

Fix: extract raw field values once per field using getAs before the function loop, storing them in preVals1[] and preVals2[]. For each function, call countEliminationsWithValues(preVals1, preVals2, least) which uses function.applyToValue(preVals[i]) — just the hash transformation on an already-extracted value, not a full Row field lookup. Add early exit: once elimCount >= least (current best), stop counting rows immediately — this candidate cannot win. Defer buildDupeRemaining to after all candidates are evaluated, calling it once only for the winner.

Combined impact: (1) + (2) together eliminate the biggest source of garbage in the hot loop.
All three are single-threaded improvements — they'd stack on top of any parallelization you add later.

### DIFF
--- a/common/core/src/main/java/zingg/common/core/block/Block.java
+++ b/common/core/src/main/java/zingg/common/core/block/Block.java
@@ -23,6 +23,7 @@ public abstract class Block<D,R,C,T> implements Serializable {
 	public static final Log LOG = LogFactory.getLog(Block.class);
 	private final IHashFunctionUtility<D, R, C, T> hashFunctionUtility;
 	private FieldDefinitionStrategy<R> fieldDefinitionStrategy;
+	private GetBestNode<D, R, C, T> getBestNodeStrategy = new GetBestNode<>();
 
 	protected ZFrame<D,R,C> dupes;
 	// Class[] types;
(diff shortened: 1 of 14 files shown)


sonalgoyal:
can you submit this against a new branch `singlethreadedblockingoptimisation`

LOGANBLUE1:
added this pr to singlethreadedblockingoptimisation

PR #1348, opened 2026-07-22
Changing test infrastructure to have different values for spark environment so that memory during test runs can be optimal  

### LINKED ISSUE #1275: configurable value of zingg properties in spark session provider in junits
current junits rely on `SparkSessionProvider` which reads properties from `zingg.properties`. `TestSparkBase`  calls this class and all our Spark junits are annotated with it. 

The `zingg.property` file has 8gb set for driver and executor. 8gb is only needed for some complex tests like e2e integration tests. 

So, we should have three classes - `SparkSessionProviderBase` which takes a properties file and creates the sessions. Then `SparkSessionProviderLite` where the default value in the file is set to 512mb or 1gb. And `SparkSessionProviderHeavy` where the value is 8gb. 

Also need two classes `TestSparkBaseLite` and `TestSparkBaseHeavy` which call the respsective spark session providers. 

All annotations of `TestSparkBase` should be moved to `TestSparkBaseLite`. Only the integration test should invoke `TestSparkBaseHeavy`

Chandanydv1234:
Closes #1275

### What & why

Our Spark JUnits all shared one `SparkSessionProvider` that read
`zingg.properties`, which sets the driver/executor to 8g. Only the end-to-end
tests actually need that — the rest were paying for 8g for no reason. This
splits the provider into a base plus two memory profiles so unit tests can run
light.

### Changes

- Added `SparkSessionProviderBase` (abstract) that holds all the session
  setup; subclasses just say which properties file to load via
  `getPropertiesFile()`.
- `SparkSessionProviderLite` loads a new `zingg-lite.properties` (1g),
  `SparkSessionProviderHeavy` loads `zingg-heavy.properties` (8g, the old file
  renamed).
- Added `TestSparkBaseLite` / `TestSparkBaseHeavy` extensions.
- Moved the 18 unit tests onto `TestSparkBaseLite`, and the two integration
  executor tests (`TestSparkExecutorsCompound`, `TestSparkExecutorsSingle`)
  onto `TestSparkBaseHeavy`.
- Removed the old `SparkSessionProvider` and `TestSparkBase`.
- Set `reuseForks=false` in spark/core's surefire config. A Spark session is a
  singleton per JVM (`getOrCreate` hands back the first one built), so without a
  fresh JVM per test class the first class to run would fix the memory for the
  rest — this keeps the 1g and 8g profiles independent.

### Tests

Ran locally with `mvn -pl spark/core -am test`. The provider logs
`Spark driver memory: X GB` on startup, so the numbers below are what each
session actually came up with:

- `test-compile` for spark/core + upstream modules — compiles clean.
- `TestSparkFrame` (Lite) — 1 GB — 32 run, 0 failures, 2 skipped.
- `TestSparkExecutorsCompound` (Heavy) — 8 GB — 1 run, 0 failures.
- `TestSparkExecutorsSingle` (Heavy) — 8 GB — 1 run, 0 failures.
- Sanity-checked the original code behaves the same, so no regression.

### One open question

The issue says "only the integration test" for Heavy, but I put both
`TestSparkExecutorsCompound` and `TestSparkExecutorsSingle` on Heavy since both
run full pipelines off `configSparkIntTest.json`. Happy to move `Single` back to
Lite if you'd rather stick to the letter of the issue.


sonalgoyal (changes requested):
instead of a static getInstance method, can you have a static attribute named isSessionInitialised just like https://www.baeldung.com/java-junit5-run-before-all-classes ?

Chandanydv1234:
> instead of a static getInstance method, can you have a static attribute named isSessionInitialised just like https://www.baeldung.com/java-junit5-run-before-all-classes ?

Done as suggested

PR #1355, opened 2026-08-04
Publish performance charts on the README

Chandanydv1234:
Adds a workflow that regenerates the performance history charts and publishes them to the README, keeping them current automatically.

## Changes
- New workflow `perf-charts.yml`: runs after either performance test completes (and manually via workflow_dispatch), regenerates the charts, and commits them under `assets/`.
- README: new Performance section showing both dataset charts.

## Notes
- Reuses the existing `GH_PERF_COMMUNITY_ACTION` secret.
- Depends on the updated `plot_result.py` in zingg_performance `main` (zinggAI/zingg_performance#9).

### DIFF
--- a/.github/workflows/perf-charts.yml
+++ b/.github/workflows/perf-charts.yml
@@ -1,0 +1,62 @@
+name: performance-charts
+
+# Regenerates the performance history charts from the committed test-report
+# history and publishes them into this repo (shown on the README). Runs after
+# either performance test completes, and can also be triggered manually.
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows:
+      - performance-test-febrl120K
+      - performance-test-ncVoters5M
+    types:
+      - completed
+
+permissions:
+  contents: write
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/perf-charts.yml
```diff
@@ -0,0 +1,62 @@
...
+        run: pip install pandas matplotlib
+
+      - name: Clone performance reports repo
+        run: |
+          set -e
+          git clone https://x-access-token:${{ secrets.GH_PERF_COMMUNITY_ACTION }}@github.com/zinggAI/zingg_community_performance_reports.git
+
+      - name: Fetch chart script
+        run: |
+          wget -O plot_result.py https://raw.githubusercontent.com/zinggAI/zingg_performance/main/plot_result.py
```
sonalgoyal, line 41: we dont want to commit the charts, we want to only commit the raw data, charts should only be displayed

sonalgoyal (changes requested):

PR #1356, opened 2026-08-05
Fix labeler truncating field values in vertical display

Chandanydv1234:
The labeler rendered pairs via ZFrame.show(int), which lets Spark default to truncate=true, capping cell values at 20 characters (e.g. emails shown as john_smith@cas...). Add a show(int, boolean) overload to ZFrame and SparkFrame, and have VerticalDisplayUtility call show(MAX_COLUMNS, false) so full values are displayed.


padam-prakash (approved):
Looks fine

padam-prakash:
reverted and merged in main with #1361

PR #1364, opened 2026-08-11
updated jackson version

LOGANBLUE1:


sonalgoyal (changes requested):
have you verified these changes do not affect enterprise snowflake? We do not want a snowflake declaration if we are not importing this version.

LOGANBLUE1:
there is no dependency change in this PR, so no breaking changes.

PR #1377, opened 2026-08-17
Add SparkDocumenter executor test (#1304)

### LINKED ISSUE #1304: Create Spark Documeter testcases
There is no testcase for spark documeter

Chandanydv1234:
## What

Adds an executor-level test for the `SparkDocumenter` phase (`generateDocs`). Until now the documenter component classes (`ModelDocumenter`, `DataDocumenter`, `DocumenterBase`) had unit tests, but the executor's `Documenter.execute()` path — which wires them together — had no coverage. Every other phase (Trainer, Matcher, Labeller, Linker) has an executor test; Documenter was the gap.

## How

- `TestDocumenterExecutorBase` — engine-agnostic base test. Loads the documenter test config, points `zinggDir` at the prepared model-100 fixture, points the input data at the test CSV, runs `execute()`, and asserts both `model.html` and `data.html` are generated.
- `TestSparkDocumenterExecutor` — Spark binding via `TestSparkBaseLite`.

Generated documents are deleted in a `finally` block so they never pollute the source tree.

## Testing

All documenter tests pass locally (JDK 11, Spark 3.5):

```
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
```

Closes #1304

### DIFF
--- a/common/core/src/test/java/zingg/common/core/documenter/TestDocumenterExecutorBase.java
+++ b/common/core/src/test/java/zingg/common/core/documenter/TestDocumenterExecutorBase.java
@@ -1,0 +1,85 @@
+package zingg.common.core.documenter;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.rmi.NoSuchObjectException;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import zingg.common.client.ClientOptions;
+import zingg.common.client.arguments.ArgumentServiceImpl;
+import zingg.common.client.arguments.IArgumentService;
+import zingg.common.client.arguments.model.Arguments;
+import zingg.common.client.arguments.model.IArguments;
+import zingg.common.client.pipe.FilePipe;
+import zingg.common.client.pipe.Pipe;
+import zingg.common.client.ZinggClientException;
+import zingg.common.core.context.Context;
+import zingg.common.core.executor.Documenter;
+
+/**
+ * Tests the Documenter executor end to end: running {@link Documenter#execute()}
+ * against a prepared model fixture should produce both the model and data HTML
+ * documents on disk.
+ */
+public abstract class TestDocumenterExecutorBase<S, D, R, C, T> {
+
+    public static final Log LOG = LogFactory.getLog(TestDocumenterExecutorBase.class);
+
+    protected Context<S, D, R, C, T> context;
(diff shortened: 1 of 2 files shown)


common/core/src/test/java/zingg/common/core/documenter/TestDocumenterExecutorBase.java
```diff
@@ -0,0 +1,85 @@
+package zingg.common.core.documenter;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.rmi.NoSuchObjectException;
```
sonalgoyal, line 7: why are we adding rmi imports?

common/core/src/test/java/zingg/common/core/documenter/TestDocumenterExecutorBase.java
```diff
@@ -0,0 +1,85 @@
...
+        this.context = context;
+    }
+
+    /** each engine (Spark, etc.) supplies its own initialised executor */
+    protected abstract Documenter<S, D, R, C, T> getDocumenter() throws ZinggClientException;
+
+    @BeforeEach
+    public void setUp() throws NoSuchObjectException, ZinggClientException {
+        String configPath = getClass().getResource("../../../../documenter/config.json").getFile();
+        IArgumentService<Arguments> argsUtil = new ArgumentServiceImpl<>(Arguments.class);
```
sonalgoyal, line 50: create arguments programmatically. talk with @LOGANBLUE1 about this

sonalgoyal (changes requested):
documenter in enterprise snowflake writes to tables. the current code hard codes the locations/paths of the documentation, we need to abstract them. 

please also submit corresponding pr in enterprise

common/core/src/test/java/zingg/common/core/documenter/TestDocumenterExecutorBase.java
Chandanydv1234: understood

common/core/src/test/java/zingg/common/core/documenter/TestDocumenterExecutorBase.java
Chandanydv1234: To stay consistent I borrowed TestModelDocumenterBase and TestDataDocumenterBase as templates. So it is the same reason they do. setup() calls loadArguments(), which can throw NoSuchObjectException, so setup() has to throw the same error. and this file imports it

sonalgoyal: please change that as well

padam-prakash:
check if we can add field checks in the created doc files ?

PR #1391, opened 2026-09-08
added config changes

LOGANBLUE1:


padam-prakash (approved):
Approved

common/core/src/test/java/zingg/common/core/executor/ExecutorTester.java
```diff
@@ -23,39 +18,44 @@ public class ExecutorTester<S, D, R, C, T>{
...
-		return args;
 	}
 
-	public void setupArgs() throws ZinggClientException, IOException{
-		this.args = setupArgs(configFile, "");
+	/**
+	 * Hook for testers that need to adjust args before the run, e.g. attach an in-memory
+	 * dataset to a pipe. Does nothing by default.
+	 */
+	protected void updateArgs(IArguments args) throws Exception {
```
padam-prakash, line 58: can we remove this and override setupArgs() in enterprise for appending modifications or move setmodeid() into update args which get overridden in the enterprise?

